PR #4, opened 2023-01-07
Adds Lando for local development 

ChrisMKindred:
## What does this do/fix?

Adds the ability to run with [Lando](https://lando.dev/).
Updates the `readme.md` for details on backend setup, and local environments.

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -1,0 +1,29 @@
+name: blocks-starter
+recipe: wordpress
+config:
+  php: '8.0'
+  database: mariadb
+  composer_version: 2
+  via: nginx
+  xdebug: true
+  memcached: true
+services:
+  database:
+    portforward: 3306
+  mailhog:
+    type: mailhog
+    portforward: true
+    hogfrom:
+      - appserver_nginx
+      - appserver
+tooling:
+  xdebug-on:
+    service: appserver
+    description: Enable xdebug for nginx.
+    cmd: docker-php-ext-enable xdebug && pkill -o -USR2 php-fpm && echo "Enabling xdebug enabled"
+    user: root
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1,20 +1,39 @@
...
+
+``` bash
+composer setup-project
+# ... update .env file if you need ACF Pro and Gravity Forms
+composer install
+composer setup-wordpress
+```
+
+For WordPress updates, you can change the `--version=` value in the `setup-wordpress` composer script.
+
```
dpellenwood, line 36: Suggestion: It might be good to move this up before the Local Development section. I got caught up following the readme in-order and started up Lando before running the composer steps.

dpellenwood (approved):
Ran through this again this morning and it worked like a charm. Just one thought about readme order.

Also, I think we should avoid the Safe-SVG plugin. IIRC, [Jonathan found that it was super inefficient and causes issues with resizing/built-in WordPress features.](https://github.com/moderntribe/tribe-libs/compare/93bc0ecf5bc2...91f7149c13e7) Hence our [SVG support in Tribe Libs](https://github.com/moderntribe/tribe-libs/tree/4.x/src/Media/Svg).

.lando.yml
```diff
@@ -0,0 +1,29 @@
+name: blocks-starter
```
dpellenwood, line 1: Moose?

README.md
ChrisMKindred: You should run this with Lando started. If you are running Lando you need to run `lando composer [command]` in order to run the command in the container... 

With that being said, I'll add a quick explainer for both Local and Lando on making sure you run the composer commands in their respective containers.

PR #6, opened 2023-02-13
Add dependent workflows

tarecord:
## What does this do/fix?

This creates some dependent workflows to fail fast/often to conserve Action minutes for unnecessary workflow runs. For example, if there are issues with coding standards, we should fail quickly and wait to run the more intensive workflows until issues are fixed.

The Actions should run in this order and failing steps will halt workflows further down the chain:
1. [Coding Standards](https://github.com/moderntribe/moose/blob/8018cd405916c58d745ae0994d38e418338b2de6/.github/workflows/phpcs.yml)
2. [Static Analysis](https://github.com/moderntribe/moose/blob/8018cd405916c58d745ae0994d38e418338b2de6/.github/workflows/phpstan.yml)
3. [PHP Compatibility](https://github.com/moderntribe/moose/blob/8018cd405916c58d745ae0994d38e418338b2de6/.github/workflows/compatibility.yml)
4. [Slic Automated Tests](https://github.com/moderntribe/moose/blob/8018cd405916c58d745ae0994d38e418338b2de6/.github/workflows/slic.yml)

Screenshot on 2023-02-13 at 11-03-16

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -1,0 +1,38 @@
+name: PHP Compatibility
+
+on: ['workflow_call']
+
+jobs:
+  php-compatibility:
+    name: PHP ${{ matrix.php-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-version: ["7.4", "8.0", "8.1", "8.2"]
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Configure PHP environment
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.4
+          extensions: mbstring, intl
+          coverage: none
+
+      - name: Get Composer Cache Directory
+        id: composer-cache
+        run: |
+          echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
+      - uses: actions/cache@v3
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-composer-
+      - uses: ramsey/composer-install@v2
+        with:
+          dependency-versions: highest
+
+      - name: Run PHP Compatibility
+        run: composer compatibility:php-${{ matrix.php-version }}
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -1,0 +1,20 @@
+name: Coding Standards
+
+on: ['workflow_call']
+
+jobs:
+  phpcs:
+    name: phpcs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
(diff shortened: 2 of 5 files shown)


.github/workflows/compatibility.yml
```diff
@@ -0,0 +1,38 @@
...
+
+on: ['workflow_call']
+
+jobs:
+  php-compatibility:
+    name: PHP ${{ matrix.php-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-version: ["7.4", "8.0", "8.1", "8.2"]
```
ChrisMKindred, line 11: I am glad we are able to get all of these into a compatibility test.  Can we set this up to throw a warning/notification on "8.2" but not fail?

.github/workflows/phpcs.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: phpcs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Configure PHP environment
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.0'
+      - uses: ramsey/composer-install@v2
```
ChrisMKindred, line 16: This needs to also get the .env secrets, or we will never get past the `composer install`.

.github/workflows/slic.yml
```diff
@@ -0,0 +1,76 @@
...
+          ${SLIC_BIN} xdebug off
+          ${SLIC_BIN} debug on
+          ${SLIC_BIN} info
+          ${SLIC_BIN} config
+      - name: Set up Moose
+        run: |
+          ${SLIC_BIN} use moose
+          ${SLIC_BIN} composer install --ignore-platform-reqs
+      - name: Run suite ${{ matrix.suite }}
+        run: ${SLIC_BIN} run ${{ matrix.suite }} --ext DotReporter
```
ChrisMKindred, line 76: We should setup a quick example test allowing this to run.

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Configure PHP environment
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.0'
+      - uses: ramsey/composer-install@v2
+        with:
+          composer-options: "--ignore-platform-reqs --optimize-autoloader"
```
ChrisMKindred, line 18: if we are running 8.0 here, do we need the `--ignore-platform-req`?

.github/workflows/phpcs.yml
```diff
@@ -0,0 +1,20 @@
...
+        with:
+          composer-options: "--ignore-platform-reqs --optimize-autoloader"
```
ChrisMKindred, line 18: if we are running 8.0 here, do we need the `--ignore-platform-req`?

ChrisMKindred (commented):
This is great!  I have a couple of questions/comments for you before approving.

PR #9, opened 2023-03-13
Adds an index.php file

ChrisMKindred:
## What does this do/fix?

Updates the theme to have an `index.php`
Updates the Theme Name to Moose because I want to stay away from using SquareTwo

## QA

Screenshots/video:
- [Error when loading](http://p.tri.be/i/59Bvmr)

### DIFF
--- a/wp-content/themes/core/index.php
+++ b/wp-content/themes/core/index.php
@@ -1,0 +1,3 @@
+<?php declare(strict_types=1);
+
+// Silence is golden
--- a/wp-content/themes/core/style.css
+++ b/wp-content/themes/core/style.css
@@ -1,5 +1,5 @@
 /*
-Theme Name: SquareTwo
+Theme Name: Moose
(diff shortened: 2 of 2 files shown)


dpellenwood:
Hey there, I'd like to hold on this one for a bit (A week or two) while I work with Sarah to define the baseline theme.

For context, I 🧇 between adding this or an [index.html per FSE](https://fullsiteediting.com/lessons/creating-block-based-themes/#h-step-2-add-your-first-block-template).

ChrisMKindred:
I get that. For context, it is a broken theme when you pull down and build the project as is. I would rather at least have it working, and then when that decision is made, we can update it.

dpellenwood (approved):
> I get that. It is a broken theme when you pull down and build the project as is. I would rather at least have it working, and then when that decision is made, we can update it.

Makes sense. Let's merge this then. Thanks!

PR #10, opened 2023-03-27
Force Plugin Activation Refactor to 2.0

ChrisMKindred:
## What does this do/fix?

There was a refactor by StellarWP on the Force Plugin Activation. This brings that functionality into the plugin using our plugin list.


dpellenwood (approved):
Love it. Merge away!

PR #13, opened 2023-04-04
Adds Testing Framework

ChrisMKindred:
## What does this do/fix?

Adds a basic testing framework.
Sets up dependent actions.


composer.json
```diff
@@ -89,10 +89,18 @@
     "require": {
         "php": "^7.4|^8.0",
         "advanced-custom-fields/advanced-custom-fields-pro": "*",
+        "codeception/module-asserts": "^1.0",
+        "codeception/module-cli": "^1.0",
+        "codeception/module-db": "^1.0",
+        "codeception/module-filesystem": "^1.0",
+        "codeception/module-phpbrowser": "^1.0",
+        "codeception/module-webdriver": "^1.0",
+        "codeception/util-universalframework": "^1.0",
```
dpellenwood, line 98: Should these be in `require-dev` instead of `require`? (Same with wp-browser, below?)

dpellenwood (approved):
One question, but looks good overall.

PR #14, opened 2023-04-12
[MOOSE-8] Styleguide -  Color

jamiepaul:
## What does this do/fix?

Styleguide updates:
- Adds color and shadow settings to `theme.json`
- Adds color `_variables.pcss` based on Figma Styleguide
-- Associates w/ theme.json WP presets where relevant
-- Assumes `theme.json` color slugs in theme.json will be updated once we know more about how the colors from the styleguide will be used (foreground/background)

Architectural updates:
- Removes postcss-global-data dependency
- Updates directory structure under theme's assets/pcss/ similar to SQ1 where `_variables.pcss` partials are split out and organized by context
- Adds templates directory w/ a simple `page.html` template to see page content temporarily until a template architecture strategy is solidified
- Adds header/footer as template parts as placeholders 

## QA

Links to relevant issues


wp-content/themes/core/theme.json
```diff
@@ -1,10 +1,77 @@
...
+    "color": {
+      "custom": false,
+      "customDuotone": false,
+      "customGradient": false,
+      "defaultGradients": false,
+      "defaultPalette": false,
+      "duotone": [],
+      "gradients": [],
+      "link": false,
+      "palette": [
```
jamiepaul, line 14: Note: I didn't define all colors from the styleguide here as I think we'll want a curated list based on what should be available for content editors. I assume as we get further into core block styles, color theming, etc we'll be able to hone these in.

wp-content/themes/core/assets/pcss/global/color/_variables.pcss
```diff
@@ -0,0 +1,86 @@
...
+	 * Brand: Secondary
+	 * ----------------------------------------------------------------------------- */
+
+	--color-brand-secondary: var(--wp--preset--color--brand-secondary);
+
+	/* -----------------------------------------------------------------------------
+	 * Neutral: Grey
+	 * ----------------------------------------------------------------------------- */
+
+	--color-neutral-10: #f5f5f5;
```
jamiepaul, line 28: Note: I'd like to tie as many color properties here to the theme.json color palette variables as possible so they're connected, but I think it's best to wait and see how the design shakes out further to figure out a strategy for how they should be connected.

wp-content/themes/core/assets/pcss/global/color/_variables.pcss
```diff
@@ -0,0 +1,86 @@
...
+	--color-brand-primary: var(--wp--preset--color--brand-primary);
+
+	/* -----------------------------------------------------------------------------
+	 * Brand: Secondary
+	 * ----------------------------------------------------------------------------- */
+
+	--color-brand-secondary: var(--wp--preset--color--brand-secondary);
+
+	/* -----------------------------------------------------------------------------
+	 * Neutral: Grey
```
jamiepaul, line 25: Note: I'd like to tie as many color properties here to the theme.json color palette variables as possible so they're connected, but I think it's best to wait and see how the design shakes out further to figure out a strategy for how they should be connected.

wp-content/themes/core/theme.json
```diff
@@ -1,10 +1,77 @@
...
+  "settings": {
+    "color": {
+      "custom": false,
+      "customDuotone": false,
+      "customGradient": false,
+      "defaultGradients": false,
+      "defaultPalette": false,
+      "duotone": [],
+      "gradients": [],
+      "link": false,
```
jamiepaul, line 13: Note: I disabled link color as I think historically our design system has handled this and it's not likely to change

wp-content/themes/core/theme.json
```diff
@@ -1,10 +1,77 @@
...
-	"settings": {
-		"layout": {
-			"contentSize": "80ch",
-			"wideSize": "1200px"
-		}
-	}
+  "$schema": "https://schemas.wp.org/trunk/theme.json",
+  "version": 2,
+  "settings": {
+    "color": {
```
jamiepaul, line 5: ❓ I left background color settings enabled. We disabled it for LendNation. Do we want it disabled here to start?

package.json
```diff
@@ -27,7 +27,6 @@
     "coreThemeBlocksDir": "./wp-content/themes/core/blocks"
   },
   "devDependencies": {
-    "@csstools/postcss-global-data": "^1.0.3",
```
dpellenwood, line 30: Ah, 💩 . I must have made this unclear in my Slcak convo. I'm hoping to still use this plugin, but not using a `/globals/` directory for items to include. Instead, we can reference the variable & mixin partials across any number of directories using this plugin.

So, for example, we could still set up the color partials as you have already, but we can add `${ pkg.config.coreThemeDir }/assets/pcss/**/_variables.pcss` or maybe we should be more specific to only the partials we want to use.

wp-content/themes/core/assets/pcss/_variables.pcss
```diff
@@ -0,0 +1,17 @@
...
+ * ----------------------------------------------------------------------------- */
+
+/* Colors */
+@import "global/color/_variables.pcss";
+
+/* Custom Selectors */
+@import "global/custom-selectors/_variables.pcss";
+
+/* Media Queries */
+@import "global/media-queries/_variables.pcss";
```
dpellenwood, line 17: Using the global imports plugin eliminates the need for us to do this, and it provides access to all these partials when we do block-specific entrypoints. :)

package.json
dpellenwood: Additionally, Imho, I think we can skip the `global` directory all together. ANy styles specific to a component, block, pattern, or template should live in the same directory as the markup & other respective files for that item. (Similar to Sq1's components.)

wp-content/themes/core/theme.json
dpellenwood: I'd say yes for now.

wp-content/themes/core/theme.json
dpellenwood: I say we define them all here. We can tailor a subset of colors available for different blocks/contexts, but I'd prefer to have everything available from the json file initially.

Additionally, I think this global palette should be named 1:1 with what is defined in [the global palette in Figma](https://www.figma.com/file/zbjC33tZJz5A901jeXM7zv/Website-Design-System?node-id=804-5786&t=5pO9WEjMQHEbgwrv-0) (same naming even). Once we define contextual palettes for different blocks/use cases, we can add the relevant names (foreground, text, etc).

wp-content/themes/core/assets/pcss/print.pcss
```diff
@@ -1,11 +1,5 @@
...
+/*
+ * Print CSS
+ */
 
-.theme {
-
-	/* Some theme postcss */
-	color: var(--color-primary);
-}
+/* TODO: Bring over print.pcss from SQ1? */
```
dpellenwood, line 5: Good catch. [Ticketed.](https://moderntribe.atlassian.net/browse/MOOSE-20)

dpellenwood (changes requested):
This is a great start, Jamie. Just a couple architectural items and a few smaller suggestions. Sorry for the confusion re: the global data plugin. Lemme know if you want to chat that through more.

package.json
jamiepaul: @dpellenwood Gotcha, that makes sense re: use of the globals postcss plugin. I'll update.

What are your thoughts on renaming /assets/pcss/**global**/ to /assets/pcss/**theme**/? That directory would hold things like color, custom selectors, media queries, type mixins, etc.. things that need to be setup for the general theme. It wouldn't include styles specific to blocks/components.

On [LendNation](https://github.com/wpcomvip/qchi-lendnation/tree/develop/themes/core/assets/pcss), we needed to include `/patterns/*.pcss` and `/templates/*.pcss` at the same level as `/assets/pcss/global` because we didn't have a way to load style partials for patterns or templates individually like blocks. I assumed we'd need to do the same here, so the structure would be something like:
```
pcss/
- theme/
--- color
--- custom-selectors
--- media-queries
--- typography

- patterns/
--- hero

- templates/
--- 404
--- single 
```

wp-content/themes/core/theme.json
jamiepaul: Makes sense, I like that, will update 👍

package.json
dpellenwood: I think we can avoid global and/or theme. as the css in the assets directory will be only that. Totally get how it was set up on LendNation. We're planning to make it possible to enqueue the css for templates/patterns/components from within those respective directories here. There's a backend ticket for that in the works already. :)

wp-content/themes/core/theme.json
jamiepaul: ✔️ These have been updated

wp-content/themes/core/assets/pcss/_variables.pcss
jamiepaul: ✔️ This file is gone now

package.json
dpellenwood: You configured the plugin correctly. any css defined in a `_variables.pcss` file will now be available to use in any other processed pcss file, even if the variables file(s) were not specifically imported, yes.

dpellenwood (approved):
Looks great! Thanks for tackling this.

PR #15, opened 2023-04-13
[MOOSE-8] Styleguide typography

jamiepaul:
## What does this do/fix?

- Adds Google typeface "Inter" normal and bold font files (separate files for italic at each weight aren't provided for this typeface)
- Adds fontSize scale to theme.json without fluid typography turned on (we want to calculate our own clamp values, as design didn't love WP's generated clamps)
- Adds default element typography styles for headings, caption and cite as well as the paragraph block in theme.json
- Adds pcss/typography directory w/ variables tied to WP presets
- Adds mixins for heading and body text styles, along with utility classes
- Adds `postcss-mixins` devDependency and configures `postcss-global-data` to capture `_mixins.pcss` partials from the assets/pcss/ directory

## QA

Links to relevant issues
- [Link to MOOSE-8](https://moderntribe.atlassian.net/browse/MOOSE-8)

Screenshots/video:
image
image


wp-content/themes/core/theme.json
```diff
@@ -95,10 +95,100 @@
         }
       ]
     },
+    "custom": {
+      "fontWeight": {
```
jamiepaul, line 99: I added these as custom WP presets because I've seen different typefaces have differing numeric values for font-weight names. For example, one font will define bold 800 and different font will define it as 700. 

I could be overcomplicating it, but variables help me not have to remember the specific project's font-weight values.

wp-content/themes/core/theme.json
```diff
@@ -95,10 +95,100 @@
...
+        "regular": 400,
+        "bold": 700
+      }
+    },
     "layout": {
-      "contentSize": "80ch",
+      "contentSize": "680px",
```
jamiepaul, line 105: Not related to styleguide typography but updated to a px value here so blocks align appropriately. Will update this again later when spacing and layout styleguide aspects are implemented.

wp-content/themes/core/theme.json
```diff
@@ -95,10 +95,100 @@
...
       "wideSize": "1200px"
     },
+    "typography": {
+      "customFontSize": false,
+      "dropCap": false,
```
jamiepaul, line 110: Death to dropcap!

wp-content/themes/core/theme.json
```diff
@@ -95,10 +95,100 @@
...
+      "fluid": false,
```
jamiepaul, line 111: WP's fluid typography is disabled because we want more control over the clamp values

wp-content/themes/core/theme.json
```diff
@@ -95,21 +95,111 @@
...
+          "slug": "default",
+          "shadow": "0px 0px 1px rgba(48, 49, 51, 0.05), 0px 8px 16px rgba(48, 49, 51, 0.1)"
         },
         {
-          "name": "Deep",
-          "slug": "deep",
-          "shadow": "12px 12px 50px rgba(0, 0, 0, 0.4)"
+          "name": "Hover",
+          "slug": "hover",
+          "shadow": "0px 0px 1px rgba(48, 49, 51, 0.05), 0px 8px 50px rgba(48, 49, 51, 0.1)"
```
jamiepaul, line 202: Not related to styleguide typography, but shadow names and values were updated in Figma and it didn't deserve a whole separate PR

GeoffDusome (approved):
🟢 This looks good and makes sense!

wp-content/themes/core/assets/pcss/theme.pcss
```diff
@@ -11,8 +11,11 @@
...
 /* Variables */
 @import "color/_variables.pcss";
+@import "typography/_variables.pcss";
 
 /* Mixins */
+@import "typography/_mixins.pcss";
 
 /* Utility Classes */
 @import "color/_utilities.pcss";
+@import "typography/_utilities.pcss";
```
dpellenwood, line 21: Since we're using the global-css plugin, I don't think we need to import the variables, mixins, or utilities like this. As long as they are added in the plugin's config, they should be available for use in any css entry point (including the theme, admin, & print entries).

dpellenwood (approved):
Just one comment re: the unneeded vars, mixins & utility imports. Otherwise, looks great!

wp-content/themes/core/assets/pcss/theme.pcss
jamiepaul: Without importing, the theme entrypoint doesn't have access to the variables. I know the path is right in `postcss.config.js` because it was giving me errors before.. I'll have to dig a bit deeper into the configuration. I wish that plugin had better docs.

jamiepaul: @dpellenwood I spent a half hour trying a couple different things with configuration, paths, and trying to disable preserving css properties with no luck.

I'm not at all familiar w/ how the build system works, how our `postcss.config.js` hooks in to WP scripts and how to modify the order of pcss plugins. If you've got time to explain it to me like I'm 5 or point me in the direction of docs you've found useful, I'd appreciate it!

In the mean time, mind if I merge this and we can put up a fix separately?

dpellenwood (approved):
I think we're in a good spot here, @jamiepaul . Feel free to merge this as you have time. Thanks!

PR #23, opened 2023-04-19
Chore/updates

dpellenwood:
## What does this do/fix?
* Updates composer deps including ACF pro to latest.
* Updates Node & NPM to latest LTS versions. (This was held back previously b/c `wp-scripts` didn't support newer versions.)


jamiepaul (approved):
🟢 🤌

PR #24, opened 2023-04-19
[MOOSE-8] Add light reset/normalize pcss

jamiepaul:
## What does this do/fix?

Adds light-touch reset / normalize styles inspired by [Josh Cormeau's reset](https://www.joshwcomeau.com/css/custom-css-reset/) and the latest version of [Normalize.css](https://github.com/necolas/normalize.css/)

It's possible that more form reset styles may be wanted when we begin implementing forms. Overall though, pretty impressive how much browsers have standardized and the amount of old rules I was able to remove.

Testing was done across the following:
- MacOS: Safari, Firefox, Chrome
- iOS: Safari, Chrome
- Android (Samsung Galaxy devices): Chrome, Firefox, system browser

## QA

Links to relevant issues
- [Link to Issue](https://example.com)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ wp-content/wp-cache-config.php
 wp-content/tests/
 wp-content/logs/
 wp-content/mu-plugins/mu.local.php
+wp-content/mu-plugins/local-by-flywheel-live-link-helper.php
 
 wp-content/themes/*
 wp-content/plugins/*
--- a/wp-content/themes/core/assets/pcss/global/reset.pcss
+++ b/wp-content/themes/core/assets/pcss/global/reset.pcss
@@ -1,0 +1,76 @@
+/* -------------------------------------------------------------------------
+ *
+ * Global "Resets" & Light Normalize
+ *
+ * ------------------------------------------------------------------------- */
+
+/**
+ * Use a more-intuitive box-sizing model.
+ */
+*,
+*::before,
+*::after {
+	box-sizing: border-box;
+}
+
+/**
+ * Prevent adjustments of font size after orientation changes in iOS.
+ */
+html {
+	text-size-adjust: 100%;
+}
+
+/**
+ * 1. Improve text rendering.
+ * 2. Prevent horizontal scroll.
+ */
+body {
+	-webkit-font-smoothing: antialiased;
+	overflow-x: hidden;
+}
+
+/**
+ * Avoid text overflows.
+ */
+p,
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+	overflow-wrap: break-word;
+	hyphens: auto;
+}
+
+/**
+ * Add the correct font size in all browsers.
+ */
+small {
+	font-size: 80%;
+}
+
+img,
+picture {
+	height: auto;
+	max-width: 100%;
+}
+
+iframe,
+video,
+embed {
+	max-width: 100%;
+}
+
+/**
+ * 1. Remove built-in form typography styles and prevent zoom on mobile devices.
+ * 2. Remove margin in Firefox & Safari.
+ */
+input,
+button,
+optgroup,
+textarea,
+select {
+	font: inherit;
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -54,6 +54,7 @@ wp-content/wp-cache-config.php
 wp-content/tests/
 wp-content/logs/
 wp-content/mu-plugins/mu.local.php
+wp-content/mu-plugins/local-by-flywheel-live-link-helper.php
```
jamiepaul, line 57: This is auto-generated when using Local by Flywheel for a local environment and Live Links is enabled. I've used this to speed up testing across devices IRL and w/ Browserstack.

wp-content/themes/core/assets/pcss/global/reset.pcss
```diff
@@ -0,0 +1,76 @@
...
+ * 1. Improve text rendering.
+ * 2. Prevent horizontal scroll.
+ */
+body {
+	-webkit-font-smoothing: antialiased;
+	overflow-x: hidden;
+}
+
+/**
+ * Avoid text overflows.
```
jamiepaul, line 33: If we're really trying to keep things minimal, this is one thing I debated adding. Open to thoughts/opinions on this.

.gitignore
dpellenwood: Really minor, but this probably is better adde under the `# IDE generated files #` section above.

wp-content/themes/core/assets/pcss/global/reset.pcss
dpellenwood: Hmm... Yeah, I can see the hesitation. I think it's fine to add though. We end up adding it piece-meal to most project sat some point anyway.  One suggestion: can you use [the `--heading` custom selector](https://github.com/moderntribe/moose/blob/main/wp-content/themes/core/assets/pcss/custom-selectors/_variables.pcss#L13) instead of calling out all headings individually?

dpellenwood (approved):
This is great. Thanks for tackling it. A couple small suggestions and I'll be interested to get @Vinsanity & @GeoffDusome 's thoughts on the overflow-wrap.

.gitignore
jamiepaul: Good call, moved it there.

wp-content/themes/core/assets/pcss/global/reset.pcss
jamiepaul: Yes, I need to get used to those! Done.

wp-content/themes/core/assets/pcss/global/reset.pcss
```diff
@@ -0,0 +1,71 @@
...
+ */
+p,
+:--heading {
+	overflow-wrap: break-word;
+	hyphens: auto;
+}
```
GeoffDusome, line 39: I don't think you _need_ both of these here. From the docs on [overflow-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) and [hyphens](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens), it seems like you can just use `hyphens`, but `overflow-wrap` would be a good fallback if `hyphens` isn't supported for some reason.

GeoffDusome (approved):
🟢 One comment on the overflow stuff but looks good to me 👍

wp-content/themes/core/assets/pcss/global/reset.pcss
jamiepaul: Not all languages support hypens, which is why I'm using `overflow-wrap`. The `overflow-wrap: break-word` rule also looks odd in English when words just break without hypens, so I opted to use both.

There's a big more info [here](https://www.joshwcomeau.com/css/custom-css-reset/#eight-word-wrapping-9) on both rules. Appreciate you poking at this though, it's always interesting reading up on the technicalities!

Vinsanity (approved):
🟢

PR #26, opened 2023-04-20
Bugfix: Allow use of align wide and full

jamiepaul:
## What does this do/fix?

This was a fun (and frustrating) bug to track down- if “Inner blocks use content width” is disabled on Post Content then root blocks in the post editor won't have wide/full alignment controls

This fix updates the `page.html` and `single.html` templates to allow blocks supporting align wide and align full to appear in the block toolbar.

**Screenshots**:
- Without constrained layout, align options don't appear in block toolbar - http://p.tri.be/i/PV2rkF
- WITH constrained layout, align options reappear in block toolbar - http://p.tri.be/i/xZpLnu


GeoffDusome (approved):
🟢 good catch! 🌮

dpellenwood (approved):
🦍 💨

PR #27, opened 2023-04-20
[MOOSE-8] Styleguide spacing system

jamiepaul:
## What does this do/fix?

- Adds spacing scale to theme.json and map presets to CSS custom properties in `/pcss/spacing/_variables.pcss`
- Adds three aspect ratio utility classes - 1:1, 4:3 and 16:9
- Disables `blockGap` setting and applies custom styles for gap value of Buttons and Columns blocks
- Enables margin & padding in `theme.json`
- Updates layout settings `contentSize` and `wideSize`
- Adds custom WP presets for `grid-gutter` and `grid-margin`
- Enables `useRootPaddingAwareAlignments` and sets horizontal padding as grid-margin value
- Adds custom styles to ensure blocks aligned wide have the correct width

## QA

Links to relevant issues

Screenshots/video:
- [Link to Screencast explainer on the align wide custom styles needed](https://www.loom.com/share/063c7a0d84c5489a98ecd8ef11ab0b40)


wp-content/themes/core/assets/pcss/spacing/_utilities.pcss
```diff
@@ -0,0 +1,30 @@
...
+	aspect-ratio: 1 / 1;
+}
+
+.s-aspect-ratio-4-3 {
+	aspect-ratio: 4 / 3;
+}
+
+.s-aspect-ratio-16-9 {
+	aspect-ratio: 16 / 9;
+}
```
GeoffDusome, line 30: Should these be mixins pulled into helper classes so that we can use them where we don't have access to classes?

Also, I think if you put these classes on a `figure` element the `img` won't inherit the state, so the image wouldn't actually be cover but the figure would. Maybe this can be handled in the image block styling though.

wp-content/themes/core/assets/pcss/layout/default.pcss
```diff
@@ -0,0 +1,33 @@
...
+
+/*
+ * Max-width of blocks aligned wide nested inside two .has-global-padding
+ * containers need to re-account for grid-margin.
+ * Example: Columns block aligned wide inside group block aligned full.
+ */
+.has-global-padding :is(.has-global-padding) > .alignwide,
+:is(.editor-styles-wrapper) .has-global-padding :is(.has-global-padding) > .alignwide {
+	max-width: var(--wp--style--global--wide-size);
+}
```
GeoffDusome, line 33: In this case do the inner blocks even get padding? I couldn't tell from the video but it seemed like the padding had been removed in that second example and that if you responded the browser that it would be full width eventually anyway. This is mostly just me confirming that we would have padding in this situation and that we wouldn't run into responsive issues.

GeoffDusome (approved):
🟢 I think it looks good. Couple of questions but I don't think they would block anything.

wp-content/themes/core/assets/pcss/layout/default.pcss
dpellenwood, line 33: A few questions & suggestions here:
* I wonder if we need a bit more explanation of these conditions. I'm fairly comfortable with the editor, and I'm not sure I follow the complexity happening here.
* If this is caused by other styles (WP Core maybe?) is there an alternative where we unset whatever those styles are to then create the conditions we want instead?
* I wonder if we move this (or rename it) to an alignments partial and then make  `.alignwide` a top-level selector with the conditional as nested parents? I'm not sure we'll really need/have a "default" alignment system any more since the block editor technically blew that all up.

wp-content/themes/core/assets/pcss/spacing/_variables.pcss
```diff
@@ -0,0 +1,31 @@
...
+	--spacer-50: var(--wp--preset--spacing--50);
+	--spacer-60: var(--wp--preset--spacing--60);
+	--spacer-70: var(--wp--preset--spacing--70);
+
+	/* -------------------------------------------------------------
+	 * Spacers (WP Admin Bar)
+	 * ------------------------------------------------------------- */
+
+	--spacer-wpadmin-mobile: 46px;
+	--spacer-wpadmin-desktop: 32px;
```
dpellenwood, line 30: I think WordPress provides a css var for this natively now, (when the admin bar is active): `--wp-admin--admin-bar--height`. They also handle setting it properly per viewport. I wonder if we can just utilize theirs, making a comment here that there is a core-supported method for this now?

wp-content/themes/core/blocks/core/button/style.pcss
```diff
@@ -2,6 +2,10 @@
  * Styles specific to this block
  */
 
+.wp-block-buttons {
+	gap: var(--wp--preset--spacing--30);
+}
+
```
dpellenwood, line 8: Is the button**S** block technically different from the button block? I'd hate to spin up all the code needed for just this little css, but...I'd also like to make sure we're consistent with how Core divides up the blocks. Thoughts?

wp-content/themes/core/blocks/core/columns/index.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Scripts specific to this block
+ */
+
+import './style.pcss';
+import './editor.pcss';
```
dpellenwood, line 6: I wonder if we skip an editor.pcss unless it's needed?

dpellenwood (changes requested):
This is excellent progress, Jamie. I left a few questions and comments. Lemme know if you want to discuss anything further. I'm definitely open to feedback.

wp-content/themes/core/blocks/core/button/style.pcss
jamiepaul: Technically, yes, buttons vs button are different blocks.

Given the buttons block just has layout styles correlating to block settings, and I can't see a future where we'd really need to customize buttons block styles more than setting the gap value I decided to just add it in to the "button" block.

I'm all for consistencly but, it does seem like overkill (to me) to set up all the code just for the one rule. It also is easier to skim the directory and not have to mentally decide which one to open/edit if there's just the one.

wp-content/themes/core/blocks/core/columns/index.js
jamiepaul: I'm good w/ that. It doesn't appear that block editor styles are getting enqueued in the admin at all right now. If there was a way to intelligently not enqueue if the file is empty, I'd say keep them because if they're needed it's less setup. But if we're loading a bunch of empty stylesheets, then yea, lets only do it when necessary.

LendNation did end up w/ a bunch that are empty. Only a small handful (5?) needed specific editor styles.

wp-content/themes/core/assets/pcss/layout/default.pcss
GeoffDusome: @dpellenwood I added a bit more context to the comments here. I did have to play with it a bit to understand what's going on but essentially because we're using `calc(100% - var(--wp--custom--spacing--grid-margin) * 2)` as the value for `alignwide` instead of a set pixel value, we need to override cases where the padding on the left/right of blocks would be doubled. The first block is overriding one of these cases, and the second is overriding the first, essentially.

wp-content/themes/core/assets/pcss/spacing/_utilities.pcss
GeoffDusome: Updated the `.s-aspect-ratio-cover` selector to include `.s-aspect-ratio-cover > img` to also allow the `img` within the image block to cover. Otherwise, the image would take it's own form.

PR #28, opened 2023-04-20
[MOOSE-8] Focus styles

jamiepaul:
## What does this do/fix?

- Applies `:focus-visible` state as a box-shadow to ensure double border is displayed in all browsers
- Inline text that wraps lines looks better w/ outline focus style so it has a separate mixin
- CSS custom properties set up for contrasting light/dark colors used w/ box-shadow and can be updated per-project
- Applies inline text link and button style updates from Figma

## QA

Links to relevant issues

Screenshots/video:
- [Link to DQA Screencast](https://www.loom.com/share/54e491784e8e42a8aa15b274816ce21a)


dpellenwood:
> I wanted to get your opinions on implementing w/ box-shadow before I work w/ Sarah on the right color to use.
I'm good with this approach. Thanks for digging in on it.

wp-content/themes/core/assets/pcss/typography/anchors.pcss
```diff
@@ -6,6 +6,11 @@
 
 a {
```
GeoffDusome, line 7: ❓ Do we need some sort of front end selector for this so that these styles don't push out to other things? I think Dave was talking about this, if you solved this somewhere else feel free to ignore!

GeoffDusome (approved):
🟢 all of this looks really good! One minor comment.

wp-content/themes/core/assets/pcss/typography/anchors.pcss
jamiepaul: Yep, great call out. We solved for this by splitting out an `editor.pcss` file in the `/pcss/typography` directory. So, this partial isn't loaded in the editor at all.

dpellenwood (approved):
beautiful! 🐩 💨

PR #30, opened 2023-04-21
Bugfix: Theme styles in admin UI

jamiepaul:
## What does this do/fix?

Fixes an issue where global theme styles for inline text links affected WP admin UI elements.

- Adds main import files for variables, mixins and utility classes so import statements don't need to be duplicated between `/pcss/theme.pcss` & `/pcss/admin.pcss` entry points
- Creates mixin for inline text links for reuse on theme vs editor style partials
- Adds `/pcss/typography/editor.pcss` to pull in inline text link styles to the block editor and not WP admin UI elements


wp-content/themes/core/assets/pcss/typography/_mixins.pcss
```diff
@@ -104,3 +104,31 @@
 	line-height: 1.6;
 	font-weight: 400;
 }
+
+/* -------------------------------------------------------------------------
+ * Inline Links
```
jamiepaul, line 109: @dpellenwood I couldn't decide where this belongs. It felt odd to set up `/pcss/actions/_mixins.pcss` with just one mixin that relates to typography. Let me know your thoughts on architecture approach.

wp-content/themes/core/assets/pcss/_mixins.pcss
```diff
@@ -0,0 +1,14 @@
...
+ * This file is just a clearing-house.
+ * Make partials (start with an underscore) elsewhere for actual code.
+ *
+ * ------------------------------------------------------------------------- */
+
+/* Icons */
+@import "icons/_mixins.pcss";
+
+/* Typography */
+@import "typography/_mixins.pcss";
```
dpellenwood, line 14: Curious, do we even need to import mixins like this? I thought that [adding the mixinFiles setting](https://github.com/moderntribe/moose/blob/main/postcss.config.js#L58) in the postcss config would auto-locate any mixin partials in any entrypoint.

wp-content/themes/core/assets/pcss/typography/_mixins.pcss
dpellenwood: Hmm.... I do like the `actions` naming from SquareOne. Even if the partial is only a single mixin for now. It's stubbed out nicely for future additions.

dpellenwood (approved):
A couple suggestions. Thanks for tackling this!

wp-content/themes/core/assets/pcss/typography/_mixins.pcss
jamiepaul: True, I can see us wanting to use mixins for buttons/inputs and apply those to third party plugins/blocks like Gravity Forms. It is nice to have it included for future additions. I'll add.

wp-content/themes/core/assets/pcss/_mixins.pcss
jamiepaul: 🤦  I missed that, indeed you're right we don't need mixins imported. I've removed it.

PR #31, opened 2023-04-21
Bugfix: Run setup-wordpress composer command after install

ChrisMKindred:
## What does this do/fix?

This solves the problem where somebody could have updated the WordPress version in the composer file and all you have to do is run the normal composer update or composer install to make sure you are up to date with that version. 

This also solves a problem with dokku deployments needing to run the setup-wordpress command after installing.


dpellenwood (approved):
Love it.

PR #34, opened 2023-04-25
[MOOSE-23] FE: Core Blocks - Text

jamiepaul:
## What does this do/fix?

These changes apply the Figma core block styling to core WP blocks under the "text" category:
- Adds custom styles to heading block for sizes (X-Large, Large, Medium, Small, X-Small, XX-Small)
- Adds custom styles to paragraph block for sizes (Large, Small)
- Adds core list block default styles
- Adds core quote and pullquote block default styles
- Unregisters core quote block styles (default, plain)
- Adds core table block default styles
- Updates reset with additional rules to prevent browser applied spacing for some elements

Notes for future work needed:
- I've temporarily ported over the block enqueuer from LendNation so we're able to see block styles until that's solidified by BE
- We need to set default vertical spacing for elements like paragraphs, headings, lists, etc. Since the spacing PR isn't merged yet, I haven't used the spacing variables. Here's my [Figma thread](https://www.figma.com/file/zbjC33tZJz5A901jeXM7zv?node-id=1975:3703#429618879) for design to provide direction.
- The theme's JS directory architecture hasn't been discussed yet, we may want to consider what's needed and stub it out so that it can be added to easily for future projects.

## QA

Links to relevant issues
- [Link to MOOSE-23](https://moderntribe.atlassian.net/browse/MOOSE-23)

Screencast:
- [Link to Screencast Demo](https://www.loom.com/share/3dc7bdddd48c4f44b8510424106ae58a)


wp-content/themes/core/blocks/core/lists/style.pcss
```diff
@@ -0,0 +1,40 @@
...
+
+ul {
+	padding-inline-start: 0;
+}
+
+ol {
+	padding-inline-start: 1.2rem;
+}
+
+li {
```
jamiepaul, line 13: @dpellenwood something odd I noticed while styling lists- when building assets via `npm run dist` I get a [warning from postcss-nesting](http://p.tri.be/i/oQoGgu)

So, something like `ul li { ... }` would throw the warning because it doesn't start w/ a character. Seems odd since we don't have postcss-nesting as a dependency in the project. Have you seen this before? I wondered if something changed w/ the chore updates to dependencies.

jamiepaul: Dug into this a bit more- requiring nested rules to start w/ a character is part of the spec. So, maybe we want to follow that to stay as close as possible to the spec?

Related: https://chriscoyier.net/2023/03/02/nesting-in-postcss-which-way-to-go/

dpellenwood: Correct, @jamiepaul . We're following the spec as closely as possible. You can preface any element-selectors with an ampersand. Here's [a bit more clarity](https://webkit.org/blog/13813/try-css-nesting-today-in-safari-technology-preview/) on the current-version of the spec.

wp-content/themes/core/assets/pcss/admin.pcss
```diff
@@ -9,4 +9,6 @@
 @import "_utilities.pcss";
 
 /* Global Theme Editor Styles */
-@import "typography/editor.pcss";
+@import "global/reset.pcss";
+@import "typography/anchors.pcss";
```
jamiepaul, line 13: I reworked some selectors so that admin styles pick up the limited reset rules we have, and can use the same `anchors.pcss` partial as the theme stylesheet.

wp-content/themes/core/assets/js/admin.js
```diff
@@ -1 +1,12 @@
+/**
+ * unregister some block styles we don't need
+ */
+
+wp.domReady( () => {
```
jamiepaul, line 5: @dpellenwood we haven't talked much about the JS directory's architecture. I reviewed SQ1 and felt like it was a lot that we potentially don't need. I figured more discussion and a separate PR could be made for building it out, which is why this is implemented in a super basic way here.

wp-content/plugins/core/src/Assets/Block_Assets_Enqueuer.php
```diff
@@ -0,0 +1,77 @@
+<?php declare(strict_types=1);
+
+namespace Tribe\Plugin\Assets;
+
+use Tribe\Plugin\Blocks\Blocks_Definer;
+
+class Block_Assets_Enqueuer extends Assets_Enqueuer {
```
jamiepaul, line 7: @dpellenwood @ChrisMKindred @JMRhodes just an FYI that I temporarily ported this over from LendNation so I'm able to see my styles as I work through core blocks. I know BE will be looking into a better implementation soon.

GeoffDusome (approved):
🟢 All makes sense. I do like the idea of turning off default font sizes and using styles. Allows us to lock the editor into what we've defined for those sizes. I don't see any situations where we would want to allow an editor to choose a font size without also apply the styles we've defined for that font size.

Also, great job in utilizing `theme.json` for the core block styling, looks great! 🌮

wp-content/themes/core/blocks/core/lists/style.pcss
```diff
@@ -0,0 +1,47 @@
+/**
+ * Styles specific to this block
+ */
+
+/* ----------------------------------------------------------------------
+ * Ordered List
+ * ---------------------------------------------------------------------- */
+
+:where(.wp-block-post-content) ol,
```
dpellenwood, line 9: This is a nice way to handle this. I love it. I wonder if we should document somewhere all the different WP-Editor css classes that are available and where/why we're using them?

dpellenwood (approved):
This is great. Let's get it merged. @GeoffDusome , are you comfortable merging this one for us?

PR #36, opened 2023-04-26
[MOOSE-23] Core Block - Query Pagination

jamiepaul:
## What does this do/fix?

- Adds styling per Figma core blocks page design

TODO: 
- Is there a filter for the pagination-previous/next chevron setting to include our icon and write less custom styles? **GD - After some research it looks like there isn't an easy way to filter this output so I think this approach is currently the best one. We can explore further later on if needs change for the block.**
- ~~Is there a way to set defaults? Ex: Arrow setting = chevron vs none, Justification = center~~ GD - Defaults would be set in our template files.
- ~~Collab w/ design on hover/focus styles. Figma thread [here](https://www.figma.com/file/zbjC33tZJz5A901jeXM7zv?node-id=1975:3703#429886023)~~ GD - implemented these from Figma.

## QA

Links to relevant issues
- [Link to MOOSE-23](https://moderntribe.atlassian.net/browse/MOOSE-23)

Screenshots:
image


wp-content/themes/core/templates/archive.html
```diff
@@ -13,11 +13,10 @@
 		<!-- wp:post-terms {"term":"category"} /-->
 		<!-- wp:post-excerpt /-->
 	<!-- /wp:post-template -->
-	<!-- wp:query-pagination -->
-	<div class="wp-block-query-pagination">
+	<!-- wp:query-pagination {"paginationArrow":"chevron","layout":{"type":"flex","justifyContent":"center"}} -->
 		<!-- wp:query-pagination-previous /-->
+		<!-- wp:query-pagination-numbers /-->
 		<!-- wp:query-pagination-next /-->
-	</div>
```
dpellenwood, line 20: Just a thought: Could we make the query-pagination a template part and include it in each of these templates? That would keep the config & styles consistent in each template, I think?

dpellenwood (approved):
Just suggestion/question. Looks good overall.

wp-content/themes/core/templates/archive.html
GeoffDusome: Good call - I've moved the pagination blocks to a template part & updated the templates that use it. This will allow us to keep our pagination consistent across different templates.

PR #38, opened 2023-04-28
Try: Update BrowserSync Config

GeoffDusome:
## What does this do/fix?

- updates browsersync config itself, as well as the browsersync webpack plugin settings to allow the browsersync config to handle reloads.
- checks only for CSS/JS files changing in the `dist` folder. Browsersync should reload JS changes and inject CSS changes.
- checks for all PHP files in `themes/core` or `plugins/core` and will reload on any PHP save
- reorders settings on browsersync config to be alphabetical because why not


browsersync.config.js
```diff
@@ -41,10 +41,17 @@ const localConfig = moduleExists( './local-config.json' )
  */
 module.exports = {
 	debugInfo: true,
+	files: [
+		'./wp-content/themes/core/dist/**/*.js',
+		'./wp-content/themes/core/dist/**/*.css',
+		'./wp-content/themes/core/**/*.php',
+		'./wp-content/plugins/core/**/*.php',
+	],
```
dpellenwood, line 49: I wonder if this can be consolidated to:
```
		`${ pkg.config.coreThemeDir }/dist/**/*.{js,css}`,
		`${ pkg.config.coreThemeDir }/**/*.{html,php}`, / <-- Added html, since some of our patterns and templates may be html-only.
		`${ pkg.config.corePluginDir }/**/*.php`, // <-- Needs to be added to package.json
```

Note that you'd have to require the package.json file at the top of this file, too (since it's not already included):
`const pkg = require( './package.json' );`

dpellenwood (approved):
Love this improvement. One suggestion for the actual files option; but otherwise, 📦 💨

PR #39, opened 2023-05-02
[MOOSE-11] Default Pattern Categories & Clean Up

GeoffDusome:
## What does this do/fix?

- adds default pattern categories
- disables openverse media library
- disables commenting and pingbacks sitewide

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-11)


dpellenwood (approved):
Looks good to me. Couple questions for BE about how we're registering the pattern categories. I've tagged Justin R & Kris on this one, too.

wp-content/plugins/core/src/Blocks/Patterns/Pattern_Category.php
```diff
@@ -4,15 +4,23 @@
...
-			self::CUSTOM_PATTERN_CATEGORY_SLUG,
-			[
-				'label' => esc_html__( 'Custom', 'tribe' ),
-			]
-		);
+		foreach ( self::CUSTOM_PATTERN_CATEGORIES as $category ) {
+			register_block_pattern_category(
+				$category['slug'],
+				[
+					'label' => esc_html__( $category['name'], 'tribe' ),
```
ChrisMKindred, line 20: You don't want to use variables for translation.  It does work programmatically but when trying to do the translations, the software will just show `$category['nam']` and not the value from the array.

ChrisMKindred (changes requested):
There is an issue with the translation piece of the categories.  The name needs to be translated directly, not as a variable.

wp-content/plugins/core/src/Blocks/Patterns/Pattern_Category.php
GeoffDusome: I had Justin help me rewrite - should look better now!

ChrisMKindred (approved):
👍

PR #40, opened 2023-05-08
[MOOSE-8]: Styleguide Pattern

GeoffDusome:
## What does this do/fix?

- adds styleguide pattern (maybe want to get design input on the outcome / what they would want to see in this pattern)
- adds spacer block styles
- adds margin to list `li` elements by default (not editable in the editor)
- fixes small PHP error with comments code

## QA

Links to relevant issues

Screenshots/video:

Screen Shot 2023-05-08 at 14 30 07


jamiepaul (approved):
@GeoffDusome this looks great! I like that we're keeping it super basic as the starting point.

💡 My only suggestion is to add bold/italic styles somewhere in the paragraphs. Design will often check that the bold weight is set correctly and verify that italic is using the font file vs the browser's italic.

dpellenwood (approved):
Thanks for setting this up, @GeoffDusome .

PR #41, opened 2023-05-09
[MOOSE-15] Add Custom Alignments Plugin

GeoffDusome:
## What does this do/fix?

- adds custom alignments plugin to Moose
- updates composer packages

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-15)


dpellenwood (commented):
I'll leave this to Chris to do a final approval, but I'm not sure we're ready for PHP8.1 yet.

GeoffDusome:
> I'll leave this to Chris to do a final approval, but I'm not sure we're ready for PHP8.1 yet.

Looks like there's a few packages under `symfony` (`symfony/deprecation-contracts`, `symfony/filesystem`, `symfony/finder`, `symfony/process`, `symfony/string`) that require PHP 8.1. 

I am fine with reverting if we can somehow make it so that new `composer install` runs won't get us packages that require 8.1.

.lando.yml
```diff
@@ -1,7 +1,7 @@
 name: moose
 recipe: wordpress
 config:
-  php: '8.0'
+  php: '8.1'
```
ChrisMKindred, line 4: Most of the current hosts are still at 8.0. I am glad we are set up to run in 8.1, but I don't think we can default to it yet. Is it a requirement for `block-editor-custom-alignments`?

ChrisMKindred (changes requested):

.lando.yml
GeoffDusome: No - not a requirement for my plugin. Something else got updated when I reinstalled composer packages. I was getting a fatal PHP error that said 8.1 was required. See my comment above that has a bit more information.

composer.json
```diff
@@ -75,6 +75,21 @@
...
+		},
+		{
+			"type": "package",
+			"package": {
+				"name": "block-editor-custom-alignments/block-editor-custom-alignments",
+				"version": "1.0.1",
+				"type": "wordpress-plugin",
+				"dist": {
+					"type": "zip",
+					"url": "https://github.com/moderntribe/block-editor-custom-alignments/releases/download/v1.0.1/block-editor-custom-alignments.1.0.0.zip"
```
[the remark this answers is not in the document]
GeoffDusome: I updated both that version & the composer.json version before I generated the new ZIP. I wasn't sure what happened. Maybe you can take a look?

.lando.yml
ChrisMKindred: Cool, I can take a look at that.

ChrisMKindred:
@GeoffDusome I think this should be good now with those updates. Feel free to merge whenever you're ready.

ChrisMKindred (approved):

PR #42, opened 2023-05-10
[MOOSE-9] Extended Block Enqueues

JMRhodes:
## What does this do/fix?

- Adds `Block_Base` abstract class for extended core blocks to automatically register and enqueue block specific stylesheets.
- Allows blocks to define block dependencies by pre-registering block styles with a proper block handle
- Block styles are now handled in the `Block_Base` class as well.
- Added `Assets` trait, which we could further build out to help with asset paths and similar variables that we need to define.

## QA

Links to relevant issues
- https://moderntribe.atlassian.net/browse/MOOSE-9


ChrisMKindred (approved):
👍 Looks Great.

PR #43, opened 2023-05-12
Removes default support for XML-RPC Auth

ChrisMKindred:
## What does this do/fix?

Removes default support for XML-RPC authentication.  If the project requires it, we will enable it. In practice, we should have a really good reason why we need it.  Chances are really good that the Rest API can handle what is needed.


dpellenwood (approved):
Totally agree. Good catch!

PR #47, opened 2023-05-18
Bugfix/block asset path php errors

jamiepaul:
## What does this do/fix?

- Add `get_block_path()` to Query Pagination block to load block stylesheet w/ correct path
- Replace file_get_contents() usage w/ wp_remote_get() to prevent OpenSSL related PHP warnings:
```
PHP Warning:  file_get_contents(): SSL operation failed with code 1. OpenSSL Error messages:
error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed in /Users/jamie/workspace/local/moose/app/public/wp-content/plugins/core/src/Blocks/Block_Base.php on line 72
```

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "3d4dd91bbd03dc469d055960c8260704",
+    "content-hash": "d4747f7667737667cee81d0e4fd5c4f9",
     "packages": [
         {
             "name": "advanced-custom-fields/advanced-custom-fields-pro",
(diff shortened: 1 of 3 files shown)


composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "3d4dd91bbd03dc469d055960c8260704",
+    "content-hash": "d4747f7667737667cee81d0e4fd5c4f9",
```
dpellenwood, line 7: Curious why there are composer.lock changes?

dpellenwood (approved):
Just one question re: composer.lock. G2g otherwise. Thanks Jamie!

composer.lock
jamiepaul: I ran `composer update` to pull in the new block alignments plugin and other updates from when I was out. I believe anytime you run that or `composer install` the lock file gets updated.

JMRhodes (approved):
👍

composer.lock
dpellenwood: 👍  Just want to confirm that you're running composer with PHP 8, not 8.1. (We've had a couple issues with php versions + composer conflicts.)

PR #48, opened 2023-05-19
[MOOSE-12] JS architecture WIP

jamiepaul:
## What does this do/fix?

Updates:
- Built out `/assets/js/` directory structure further 
- Webpack aliases created for JS utils, config and common directories
- Refactored JS files using 3rd party libs (lodash and verge) to use vanilla JS or @wordpress packages
- Utils: added `debounce()`, removed `on()`, modernized `trigger()` and renamed `triggerCustomEvent()`
- `ready.js` for theme & admin entry are much slimmer than SQ1 - no longer init components, plugins or integrations scripts

JS directory structure outline:
- `js/admin/` includes admin entry specific scripts
- `js/theme/` includes theme entry specific scripts
- `js/config/` includes global constants, state and i18n
- `js/common/` intended for scripts used w/ both admin & theme entrypoints
- `js/utils/` includes frequently used utilities

### TODO
- Get feedback on directory structure, suggestions welcome!
- Are we localizing a JS Config object to be globally available?
- Review SQ1 JS utils in depth to determine what else should be ported

## QA

Links to relevant issues
- [Link to MOOSE-12](https://moderntribe.atlassian.net/browse/MOOSE-12)


.eslintrc.json
```diff
@@ -1,7 +1,8 @@
 {
 	"extends": [ "plugin:@wordpress/eslint-plugin/recommended" ],
 	"rules": {
-		"no-console": 0
+		"no-console": 0,
+		"import/no-unresolved": 0
```
jamiepaul, line 5: This suppresses eslint errors on import statements using the webpack aliases

wp-content/themes/core/assets/js/config/i18n.js
```diff
@@ -0,0 +1 @@
+export const NLS = window.modern_tribe_i18n || {};
```
jamiepaul, line 1: The `wp.i18n` global object wasn't always available, so I opted to just port what we had in SQ1 for now. Without actual use cases, I don't have a vision for how this could be most helpful for Moose projects. Let me know if you have ideas @dpellenwood

jamiepaul:
@dpellenwood I wanted to get this up as a WIP so you could review thus far. I'm not sure I love the directory structure I landed with.. I'm having an internal battle w/ simplifying and reducing directories vs having enough structure to build off of for actual projects.

I looked into the `wp.i18n` global object and found it wasn't always available on FE pages so I just ported what we had in SQ1 for now. Somewhat related, do you see us setting up localized JS configs?

Lastly, I haven't spent a lot of time reviewing all the util functions yet. I've done mostly redesign projects at MT, so it'd be helpful if you or @Vinsanity could chime in on which utils are most frequently used. From there I can take a look at refactoring to remove 3rd party libs and modernizing a bit.

webpack.config.js
```diff
@@ -106,6 +106,15 @@ if ( copyPluginIndex > -1 ) {
 
 module.exports = {
 	...defaultConfig,
+	resolve: {
+		...defaultConfig.resolve,
+		alias: {
+			...defaultConfig.resolve.alias,
+			utils: resolve( './wp-content/themes/core/assets/js/utils' ),
+			common: resolve( './wp-content/themes/core/assets/js/common' ),
+			config: resolve( './wp-content/themes/core/assets/js/config' ),
```
dpellenwood, line 115: wondering if we should add an alias for the theme/blocks directory?

wp-content/themes/core/assets/js/admin/editor/block-styles.js
```diff
@@ -0,0 +1,16 @@
...
+ * 				using the server-side unregister_block_styles function.
+ */
+
+const unregisterStyles = () => {
+	wp.blocks.unregisterBlockStyle( 'core/separator', 'default' );
+	wp.blocks.unregisterBlockStyle( 'core/separator', 'wide' );
+	wp.blocks.unregisterBlockStyle( 'core/separator', 'dots' );
+	wp.blocks.unregisterBlockStyle( 'core/quote', 'default' );
+	wp.blocks.unregisterBlockStyle( 'core/quote', 'plain' );
+};
```
dpellenwood, line 14: Should we move these to individual theme/blocks/core/*  blocks/component partials rather than here? I'm hoping we can keep all the scripts/styles related to a specific block in the block's directory.

wp-content/themes/core/assets/js/admin/editor/register-block-filter.js
```diff
@@ -0,0 +1,37 @@
+import { addFilter } from '@wordpress/hooks';
+
+// Specify block name and align settings for it to keep.
+const blockAlignments = {
```
dpellenwood, line 4: Similar here, It's be preferable if we had a way to set up these alignments within each block's directory rather than globally here.

wp-content/themes/core/assets/js/common/resize.js
```diff
@@ -0,0 +1,20 @@
...
+import viewportDims from './viewport-dims';
+
+const resize = () => {
+	// code for resize events can go here
+
+	viewportDims();
+
+	triggerCustomEvent( 'modern_tribe/resize_executed' );
+
+	console.log( 'Moose Admin: Resized' );
```
dpellenwood, line 17: Should this log be removed?

wp-content/themes/core/assets/js/config/i18n.js
dpellenwood: I'm pretty sure you can skip this global set up completely. If we need need to use strings in JS modules we can [import the in WP Core i18n package](https://developer.wordpress.org/block-editor/how-to-guides/internationalization/) within the specific JS module which will utilize it.

wp-content/themes/core/assets/js/utils/dom/apply-browser-classes.js
```diff
@@ -0,0 +1,33 @@
+/**
```
dpellenwood, line 1: Curious: Do you think there's a good reason to keep these browser-specific tests/classes around anymore? I can't recall the last time I utilized them for a project.

dpellenwood (commented):
Left some initial feedback. I'll follow up on the questions you outlined in the PR description. Thanks for digging in on this!

wp-content/themes/core/assets/js/admin/editor/block-styles.js
jamiepaul: I considered this and I'm not opposed. The reason I didn't implement was because we don't yet have directories for the separator or quote block, and I didn't want to create them just to unregister the block style. I looked through a few block editor projects, which block styles are commonly unregistered and believe this would be a common occurrence- needing to create the full directory structure for core blocks just for the purpose of unregistering a style.

I agree it'd be nice to have all styles/scripts related to their respective block directory so if you feel strongly we should implement that now, I can do that.

wp-content/themes/core/assets/js/utils/dom/apply-browser-classes.js
jamiepaul: I've had to use them once in the last year, so I figured having the basics set up would be nice. It's getting to be less and less though, I'd be fine taking it out and saving those extra 1-2 KBs of the JS size. Before we just remove it, thoughts on asking in the all-devs channel how often it gets used to get wider input?

webpack.config.js
jamiepaul: Yes, great callout.

wp-content/themes/core/assets/js/admin/editor/block-styles.js
dpellenwood: I'd say let's do it. It keeps the code consistent, demonstrates the approach we want to take, and makes adding additional settings, code, etc to those initial core blocks quicker on. a project.

wp-content/themes/core/assets/js/utils/dom/apply-browser-classes.js
dpellenwood: > Before we just remove it, thoughts on asking in the all-devs channel how often it gets used to get wider input?
Sounds good to me. Are you able to spin that up this week? (I'm ooo Weds  thru Mon next week.)

jamiepaul: Yea, for sure, I'll get a poll up tomorrow

webpack.config.js
```diff
@@ -61,7 +61,12 @@ const blockEntryPoints = () => {
 		{ absolute: true }
 	);
 
-	if ( ! coreBlockFiles.length ) {
+	const coreBlockAdminFiles = glob(
+		`${ pkg.config.coreThemeBlocksDir }/**/admin.js`,
+		{ absolute: true }
+	);
```
GeoffDusome, line 67: We needed another entry point for the new `admin.js` files being added for core blocks, this seemed like a good place to add this.

wp-content/plugins/core/src/Blocks/Block_Base.php
```diff
@@ -92,4 +93,30 @@ public function enqueue_block_style(): void {
...
+			true
+		);
+	}
+
+	public function enqueue_admin_scripts(): void {
+		$block = $this->get_block_handle();
+
+		wp_enqueue_script( "admin-$block-scripts" );
+	}
+
```
GeoffDusome, line 121: These functions enqueue the new `admin.js` files from `dist` with a new `file_exists` check instead of using a remote get function.

webpack.config.js
dpellenwood: Thanks for this Geoff, I'm curious why this would be necessary though? Would we be applying separate functionality against a core block in the admin versus the FE?

GeoffDusome: Are you talking about just using the `index.js` file that exists in the core block folders already and enqueuing those files instead? That feels wrong to me for some reason. I feel like having a separate `admin.js` file in the core block folders that only enqueues on the admin feels more specific and intentional on what it's used for, rather than a file that imports style files. I'm not locked to it though if you feel like it's totally unnecessary.

dpellenwood: I guess I'm more asking, JS functionality we would apply to core blocks at all. If we needed to write JS for a core block, I'd rather see us create an entirely custom block.

The reason we have an `index.js` file in the core block directories at all is to make sure that the webpack wp-scripts routine catches them as directories that need to be processed for setting up the `index.asset.php` file and compiling the imported css partials. Ideally, we would just have a `style.pcss` and/or an `editor.pcss` partial in the directory to kickoff processing, but webpack is looking for the index.js entry point to handle it's processing.

jamiepaul: @dpellenwood to answer one of your earlier questions- "Would we be applying separate functionality against a core block in the admin versus the FE?" Yes, we'd want to enqueue admin only JS for core blocks to filter block alignment support and/or disable block styles (i.e. quote default & plain styles).

Per your request [here](https://github.com/moderntribe/moose/pull/48#discussion_r1200673539) we moved the related core block admin JS to the individual block directories.

dpellenwood (approved):
🕹️ 💨

PR #49, opened 2023-05-24
[MOOSE-13] add s1 menu functions

GeoffDusome:
## What does this do/fix?

- adds menu functionality from S1 but doesn't assume anything will touch the navigation block.
- TODO: find out if the navigation block can even be filtered
- TODO: if we can't filter the navigation block what are our other options? Can we forego the navigation block entirely and just use our own custom stuff (this may be easiest since it allows us the most customization)? Can we utilize the navigation block as-is without needing extra classes/attributes/markup (if so, what would the styling structure be? per block? global?)? We'd also need to understand the limitations of the navigation block.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-13)


GeoffDusome:
Closing because this isn't needed in Moose.

PR #52, opened 2023-05-26
[TRY] Intersection Observer Animation

GeoffDusome:
## What does this do/fix?

I tried to use `is-{state}` class names for this feature as this seems like the WordPress approach to state classes. If we feel like these class names don't fit I can try something else.

- defines a new JS util for handling setting classes for elements that should animate on scroll
- uses two classes to trigger the functionality: `is-animated-on-scroll` and `is-animated-on-scroll-full`. 
- `is-animated-on-scroll` handles the functionality when the element is 25% in or out of view
- `is-animated-on-scroll-full` handles the functionality when the element is fully in view
- if an element is in view for the first time the element will get a `is-scrolled-into-view-first-time` class. This allows us to set animations on an element one time for a page load.
- If an element is intersecting it will get the `is-scrolled-into-view` class. If it is not intersecting this class will be removed.
- If an element is outside of the intersection it will receive an `is-exiting-view` class. This class can be used for exit animations.

## QA

Screenshots/video:
- [Demo Video](http://p.tri.be/v/k3lrLb)

### DIFF
--- a/wp-content/themes/core/assets/js/theme/ready.js
+++ b/wp-content/themes/core/assets/js/theme/ready.js
@@ -6,6 +6,7 @@
 
 import { ready } from 'utils/events.js';
 import { debounce } from 'utils/tools.js';
+import animateOnScroll from 'utils/animate-on-scroll.js';
 
 import resize from 'common/resize.js';
 import viewportDims from 'common/viewport-dims.js';
@@ -32,6 +33,10 @@ const init = () => {
 	// initialize global events
 
 	bindEvents();
+
+	// initialize on scroll animations
+
+	animateOnScroll();
 
 	console.info(
 		'Moose Theme: Initialized all javascript that targeted document ready.'
--- a/wp-content/themes/core/assets/js/utils/animate-on-scroll.js
+++ b/wp-content/themes/core/assets/js/utils/animate-on-scroll.js
@@ -1,0 +1,79 @@
+/**
+ * @module
+ * @exports init
+ * @description functions for handling elements that should change on scroll
+ */
+
+const el = {};
+
+/**
+ * @function handleIntersection
+ *
+ * @description Callback function for when an element comes into view
+ *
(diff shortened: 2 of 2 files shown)


dpellenwood:
Thanks for looking into this, Geoff. Curious why we'd use this versus something such as [this existing AoS package](https://github.com/michalsnik/aos)?

GeoffDusome:
> Thanks for looking into this, Geoff. Curious why we'd use this versus something such as [this existing AoS package](https://github.com/michalsnik/aos)?

I've used AoS previously and had issues with it (long pages seemed to have issues with animations further down the page). It looks like the library hasn't been updated since 2018. Maybe something else exists but I personally feel like a lightweight, browser API based solution would blatantly just be better. Not only do we have full control over the implementation, but it's just _faster_. 

The other issue I was seeing while talking through this with Jamie was that we don't necessarily have easy access to data attributes in the block editor. If we're looking specifically to incorporate this fully with the block editor, we're going to want something that's entirely class based. 

This start would allow us use a class based animation library like [animate.css](https://github.com/animate-css/animate.css) (but likely _much_ smaller) to apply animations or setup our own animations based on the classes already applied. Since each element is being watched separately, we can target each element with a different animation if necessary, or apply separate animations to elements in a section based on the container class.

I've been implementing animations this way for awhile - this is the last site (https://www.npchristian.org/) I worked on at my last job, you should be able to see a similar implementation there (sorry, it's the best example I have in action). I think this site uses mostly a component based system so the animations are just applied to each component and reused throughout the whole site. I believe animations are transitions and the base class `tbx-animated` sets `opacity: 0` on everything and then I just apply the transition when the `tbx-animated-siv-ft` class gets applied.

Ideally we would create something that allows us to set animations for elements in the block editor rather than a class based system, but this could do the job for simple scroll based animations for small budget clients. I'm not a huge fan of animations to begin with but if necessary, this can do the job.

PR #53, opened 2023-06-01
update custom alignments plugin to 1.0.4

GeoffDusome:
## What does this do/fix?

- updates block editor custom alignments plugin to 1.0.4


jamiepaul (approved):
🟢 thanks for updating!

PR #54, opened 2023-06-05
Kadence Cloud Test

ChrisMKindred:
## What does this do/fix?

Kadence Cloud and Blocks Pro Test


ChrisMKindred:
We have tested everything pretty well.  Kadence Cloud works but with the updates to WordPress, you can now create patterns in the editor which is native and does not require Kadence to work.

PR #56, opened 2023-06-26
Misc Fixes

GeoffDusome:
## What does this do/fix?

- fixes issue with core block styles not enqueuing in the admin
- fixes issue with full-width blocks blocking editors from being able to easily add blocks to the bottom of the editor window just by clicking (just add some space to the bottom of `.editor-styles-wrapper`).
- fixes issue with floated images essentially disappearing from content if using the "content size" width
- unregistered some styles from the core button block
- removes admin styles from the core button block (easy enough to add back in and we moved away from having admin files for core blocks that didn't need them)
- fixes list styles not properly pulling in because of the difference in the core block name/slug.

### DIFF
--- a/wp-content/themes/core/assets/pcss/media/_mixins.pcss
+++ b/wp-content/themes/core/assets/pcss/media/_mixins.pcss
@@ -15,7 +15,7 @@
 			max-width: 220px;
 		}
 
-		/* Fill available space left/right of contentSize block */
+		/* Ensure media can only obtain a max-width of 100% of the container */
 		@media (--viewport-large) {
 			max-width: 100%;
 		}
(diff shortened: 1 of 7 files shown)


wp-content/themes/core/assets/pcss/media/_mixins.pcss
```diff
@@ -17,7 +17,7 @@
 
 		/* Fill available space left/right of contentSize block */
 		@media (--viewport-large) {
-			max-width: calc(((100% - var(--wp--style--global--content-size)) / 2) - 2em);
+			max-width: 100%;
```
jamiepaul, line 20: @GeoffDusome does the Moose styleguide in Figma match this implementation? If not, would you leave a note for Sarah to update in Figma?

jamiepaul (approved):
🟢 thanks for getting these fixes back to Moose @GeoffDusome

dpellenwood (approved):
😍 Thanks for tackling these, Geoff!

PR #57, opened 2023-06-30
[TRY] ACF Custom Blocks Refactor

GeoffDusome:
## What does this do/fix?

- Updates the render function for custom blocks to pull from `blocks/` instead of `dist/blocks/` so that we don't need to move the template file over to `dist/`. 
- Adds example ACF block


webpack.config.js
```diff
@@ -105,20 +105,26 @@ const copyPluginIndex = defaultConfig.plugins.findIndex(
...
+			} else if ( pattern.from === '**/*.php' ) {
+				defaultConfig.plugins[ copyPluginIndex ].patterns[ index ] = {
+					from: '**/*.php',
+					noErrorOnMissing: true,
+					context: resolve( pkg.config.coreThemeDir, 'blocks/acf' ),
+					to: resolve( pkg.config.coreThemeDir, 'dist/blocks/acf' ),
+				};
+			}
+		}
+	);
```
dpellenwood, line 127: Is the underlying issue here that ACF blocks don't utilize a block.json file? Because, [ACF recommends using a block.json](https://www.advancedcustomfields.com/resources/blocks/#1-register-a-block) file for registering their blocks, just like core or other custom blocks.

wp-content/themes/core/blocks/tribe/example/block.json
```diff
@@ -1,17 +0,0 @@
...
-	"category": "tribe-custom",
-	"icon": "smiley",
-	"description": "Example block scaffolded with Create Block tool.",
-	"supports": {
-		"html": false
-	},
-	"textdomain": "example",
-	"editorScript": "file:./index.js",
-	"editorStyle": "file:./index.css",
-	"style": "file:./style-index.css"
```
dpellenwood, line 16: I'd prefer to leave the example custom block definition in Moose as we will eventually get to writing our own fully-custom blocks. :)

dpellenwood (changes requested):
Some questions & changes, if you'd be so kind. Thanks for tackling this, Geoff.

webpack.config.js
GeoffDusome: Sorry - I didn't explain this too well in the PR itself. The issue is that when we pull `block.json` over into the `dist` directory, when we set `acf.renderTemplate`, it's looking in the same folder for the file. Instead of having to write up a huge path to get back into the correct directory to read the render template, we're just pulling the render template into `dist` as well, along with the `block.json` file.

wp-content/themes/core/blocks/tribe/example/block.json
JMRhodes: In addition to this block, it'd be great to have an example block utilizing ACF. As we kickoff more of these projects, that will ideally be our go-to solution for custom blocks (at the moment). So we should ensure that all devs have a good starting-point for that.

webpack.config.js
dpellenwood: Gotcha. I'm following now. I'm fine with this approach for Careforth. I want to stew on the overall issue a bit before committing to it for Moose, if that's okay? I'm ooo next week, but as soon as I return, let's take another look at it.  Thanks!

wp-content/themes/core/blocks/tribe/example/block.json
GeoffDusome: I've added the `tribe/example` block and added an ACF example block under `blocks/acf/example` as well. 💪

GeoffDusome:
@dpellenwood We've updated this back so that we don't need to copy the PHP over to `dist`. We're now forcing the block template file to be read from the `blocks/` directory instead, which should solve some of the issues you had with our original approach. Let me know your thoughts!

dpellenwood (approved):
🦍 💨

PR #58, opened 2023-07-03
[FIX] Rename admin styles / scripts to `tribe-admin` to avoid reserved words

GeoffDusome:
## What does this do/fix?

- renames `admin` admin scripts to `tribe-admin` to avoid reserved word declarations. I think we fixed this on QCHI but it must not have made it's way back over to Moose. I'll be updating this on Careforth as well.

## QA

Links to relevant issues
- Found on Careforth in the styleguide ticket: https://moderntribe.atlassian.net/browse/SNRLNK-88


jamiepaul (approved):
🟢 🙌🏻  nice catch @GeoffDusome

PR #59, opened 2023-07-05
[MOOSE-37] 404 template

GeoffDusome:
## What does this do/fix?

- adds a simple 404 page template
- 404 page template can be edited by the client under Appearance > Editor > Templates > 404
- our template should only be used as a starting point. We can update the template in code if we feel strong enough about saving an updated version of the 404 as the "default" but it's easy enough to update across environments that I don't think it's such a big deal. 

## QA

Links to relevant issues
- [MOOSE-37](https://moderntribe.atlassian.net/browse/MOOSE-37)

Screenshots/video:
- [Screenshot](http://p.tri.be/i/JZSTrq)

[MOOSE-37]: https://moderntribe.atlassian.net/browse/MOOSE-37?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jamiepaul (approved):
🟢  this looks good to me @GeoffDusome but I noticed the ticket calls out implementing as a full page "starter pattern" so probably best to hold on merging until Dave weighs in.

I think this is a good starting point until we get going using Kadence Cloud Library for pattern management. After our first project I'm sure we'll have more opinions on how templates are implemented vs patterns.

PR #60, opened 2023-07-05
[MOOSE-33] Page Template

GeoffDusome:
## What does this do/fix?

- updates Page template to match designs
- it's probably likely that the page title element in this template would need to be removed completely on most sites. Maybe it makes sense to exclude it from the template and only have the Post Content block?

## QA

Links to relevant issues
- [MOOSE-33](https://moderntribe.atlassian.net/browse/MOOSE-33)

Screenshots/video:
- [Screenshot](http://p.tri.be/i/4jIkzh)

[MOOSE-33]: https://moderntribe.atlassian.net/browse/MOOSE-33?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/wp-content/plugins/core/src/Post_Types/Page/Config.php
+++ b/wp-content/plugins/core/src/Post_Types/Page/Config.php
@@ -15,7 +15,7 @@ class Config {
 			[
 				'core/pattern',
 				[
-					'slug' => 'patterns/page-placeholder',
+					'slug' => 'patterns/page',
 				],
 			],
 		];
--- a/wp-content/plugins/core/src/Post_Types/Page/Config.php
+++ b/wp-content/plugins/core/src/Post_Types/Page/Config.php
@@ -1,0 +1,24 @@
+<?php declare(strict_types=1);
+
+namespace Tribe\Plugin\Post_Types\Page;
+
+class Config {
+
+	protected string $post_type = Page::NAME;
+
+	/**
+	 * Define block templates for initial editor state
+	 */
+	public function register_block_template(): void {
+		$post_type_object           = get_post_type_object( Page::NAME );
+		$post_type_object->template = [
+			[
+				'core/pattern',
+				[
+					'slug' => 'patterns/page-placeholder',
+				],
+			],
+		];
+	}
+
+}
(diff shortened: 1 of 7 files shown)


jamiepaul:
> it's probably likely that the page title element in this template would need to be removed completely on most sites. Maybe it makes sense to exclude it from the template and only have the Post Content block?

@GeoffDusome @dpellenwood I agree w/ Geoff's thoughts on this. Although it's no biggie to remove the page title, it's one less thing to modify on a new project. Also, given design will be starting on patterns likely before dev gets going, removing the page title from the template allows them to just get going without anything in their way.

jamiepaul (approved):
🟢  looks good as-is, but agree we may want to just remove that page title

jamiepaul:
> Could/should we create a default page pattern which is applied by default that only injects the page title? I realize we want to use the Kadence Pattern Lib for most projects, but it would be good for us to have _some_ starting point for each template, even if it's not hard-coded in the template itself. Thoughts?

I personally don't think we need it at this stage. The page template for each project is almost always just a blocks-only page and the design varies so much per-project. I think a template for a single post that's added to the editor by default is much more likely to be useful. Curious to hear @GeoffDusome's opinion though..

GeoffDusome:
@jamiepaul @dpellenwood Removed the page title from the page template and added a new pattern, "Page Placeholder". This pattern gets added automatically to newly created pages. Likely makes sense to do this for posts as well. Simple enough to start with & easily updated with project specific blocks/patterns once we get far enough along.

wp-content/plugins/core/src/Post_Types/Page/Config.php
```diff
@@ -0,0 +1,24 @@
...
+	/**
+	 * Define block templates for initial editor state
+	 */
+	public function register_block_template(): void {
+		$post_type_object           = get_post_type_object( Page::NAME );
+		$post_type_object->template = [
+			[
+				'core/pattern',
+				[
+					'slug' => 'patterns/page-placeholder',
```
dpellenwood, line 18: Question: I wonder if we just rename this to `pattern/page` (avoid placeholder) since more than likely a dev will make modifications to the pattern as the default (rather than replacing the pattern with a different file) for each project.

dpellenwood (approved):
One suggestion re: naming of the pattern. Otherwise, good to merge!

PR #62, opened 2023-07-14
Update layout style rule comments 

jamiepaul:
## What does this do/fix?

Ran into an issue on Careforth where it wasn't clear that the layout style rules are needed only when `settings.layout.wideSize` is a fluid value, not fixed.

Moose has a fluid value for wideSize, so it makes sense to keep the rules. I've updated the comments so it's more clear they can be removed if it's a static value.

## QA

Screenshots/video:
- [Link to Screencast overview](http://p.tri.be/v/yh3QcA)

### DIFF
--- a/wp-content/themes/core/assets/pcss/layout/default.pcss
+++ b/wp-content/themes/core/assets/pcss/layout/default.pcss
@@ -13,6 +13,8 @@
  * ------------------------------------------------------------------------- */
 
 /*
+ * CASE: settings.layout.wideSize = fluid value
+ *
  * Max-width of blocks aligned wide should not account for grid margin
  * when nested in .has-global-padding because the container applies
  * grid margin as padding.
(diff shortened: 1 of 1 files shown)


dpellenwood (approved):
Nice improvement. Good to ship. I wonder honestly if the default should be a fix (not fluid) value for these?

jamiepaul:
> I wonder honestly if the default should be a fix (not fluid) value for these?

@dpellenwood I'm 50/50. For LendNation, it felt like I never had to think about gutters and spacing once we had grid container width, wide width and spacing steps set up as fluid. It felt stable, and it all just worked, which contributed to coming in under budget. However, I think WP's out of the box styles weren't built thinking people would create the align widths as a fluid values. 

Whatever is in Moose I would hope is how most designers are planning to start our projects, so might be worth bringing it up w/ Sarah before we change it.

GeoffDusome:
@dpellenwood @jamiepaul I believe the Moose styleguide has the wide size as the fluid full width with spacing on either side (http://p.tri.be/i/SHyrQX). This will likely be the starting point for most sites but Marina is aware that we have the technology to add other widths if needed so I think it's fine to leave as-is.

dpellenwood:
Thanks for the added comments, @jamiepaul  @GeoffDusome . You're good to merge here, Jamie.

PR #63, opened 2023-07-19
[MOOSE-46] Post Type Name Custom Block

GeoffDusome:
## What does this do/fix?

**NOTE: This PR is an extension of [this PR](https://github.com/moderntribe/moose/pull/61) and should be reviewed after the post cards are approved.**

- adds back in code that allows the render PHP file to be copied to the `dist/tribe/` folder. Core WP functionality copies it to `dist/`, but  we need it in the `tribe/` folder, so we have to do a bit extra to get it there.
- updates the block registrar to only use the `render_callback` args if we're talking about ACF custom blocks.
- the command to create a custom block (`npm run create-block  -- --variant dynamic`) creates the block with `.scss` files (not ideal), so, if using the command, those files will need to be renamed & updated in `index.js` / `edit.js`.
- for what we're trying to accomplish with these custom blocks, the `edit.js` file likely won't change. We're using the `ServerSideRender` package to pull in our server render of the block instead of needing to create a separate edit function that matches. 
- updated the default `textdomain` from the generated one to `tribe`. Another negative to the command I guess.
- added support for margin/padding settings on the block

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-46)

Screenshots/video:
- FE Display: http://p.tri.be/i/Q4zZ9v
- Editor Display: http://p.tri.be/i/9q4XDd


dpellenwood:
Thanks for digging into this, Geoff. I'll do a full review today yet, but wanted to mention that it looks like it's possible for us to [create & use our own templates](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-create-block/#template) for use with the create-block script, which should ease some of the pain points when using it. We could even create templates for tribe & ACF blocks, which would be pretty slick.

webpack.config.js
```diff
@@ -105,21 +105,26 @@ const copyPluginIndex = defaultConfig.plugins.findIndex(
...
+				};
+			} else if ( pattern.from === '**/*.php' ) {
+				defaultConfig.plugins[ copyPluginIndex ].patterns[ index ] = {
+					from: '**/*.php',
+					noErrorOnMissing: true,
+					context: resolve( pkg.config.coreThemeDir, 'blocks/tribe' ),
+					to: resolve( pkg.config.coreThemeDir, 'dist/blocks/tribe' ),
+				};
+			}
+		}
```
dpellenwood, line 126: I wonder if we update the render_callback template to support any php-based blocks instead of this conditional. Thoughts?

wp-content/themes/core/blocks/tribe/post-type-name/edit.js
```diff
@@ -0,0 +1,32 @@
...
+ * @see https://developer.wordpress.org/block-editor/reference-guides/block-api/block-edit-save/#edit
+ *
+ * @return {WPElement} Element to render.
+ */
+export default function Edit() {
+	const blockProps = useBlockProps();
+
+	return (
+		<div { ...blockProps }>
+			<ServerSideRender block="tribe/post-type-name" />
```
dpellenwood, line 29: Nice way to minimize the code required here. Curious: is an edit.js / `editorScript` prop required for blocks? If we omitted it, would it use the render for both contexts?

dpellenwood (approved):
A couple questions/suggestions, but I'm good with this as-is. Thanks Geoff!

wp-content/themes/core/blocks/tribe/post-type-name/edit.js
GeoffDusome: I tested out removing the `edit.js` file from `index.js` and while the block didn't break in the editor, it didn't display anything and was just a blank block. I believe the method here is doing the same thing that the `render` attribute is doing for the FE.

webpack.config.js
GeoffDusome: I did try a few ways to get these blocks working with one of them being updating the functions in `Block_Registrar.php` to try to also include the `tribe/` custom blocks and this is what I ended up landing on as the best solution. I'd be open to having BE take a look to see if they can do anything though. I've put a `@todo` in the code in the `Block_Registrar.php` file for BE to look into later. 

The main issue is that regardless of what the conditional here looks like, default WP build tools copy PHP files listed under the `render` attribute to `dist/blocks/`. So we'd either need to disable this functionality entirely (if we update `Block_Registrar.php`) or keep as-is to copy the render file to `dist/blocks/tribe`. 

ACF custom blocks don't have this issue because they use `renderTemplate`, which doesn't get picked up by the default WP build code, which is why we're easily able to update the template location.

ChrisMKindred (approved):
This looks good. As we build more custom blocks like this, I am sure we will need to keep honing our settings and maybe even creating some of our own templates for building the blocks and adding the files we need.

PR #66, opened 2023-07-20
[MOOSE-52] Filter Post Terms Block by Classes

GeoffDusome:
## What does this do/fix?

**NOTE: I am fully aware this code _sucks_. I wanted to take a stab at using the block content filter we have built into Moose (we used on M12, but haven't since then) in order to create a system for this block where we can create different version of the block via classes set on the block. I would probably want BE to take a pass at the function to see if they can clean it up.** 

- `filter-remove-links` removes the links and replaces them with `span` tags
- `filter-remove-separators` removes the separator elements from the block
- `filter-display-one-term` limits the output to one term
- any combination of the above 3 filter options can be set on block
- this solves all custom blocks we wanted to create for this issue, without needing the custom blocks. Obviously, the custom blocks would be much cleaner than this, but it does the job without needing them. It does however, force users to know that those classes exist.
- The filter doesn't work in the editor. The filter only works on the FE view of the block (this might be the biggest negative).
- The filter doesn't work with the prefix / suffix options available, I believe they just get removed.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-52)

Screenshots/video:
- Links removed and replaced with `<span>` tags: http://p.tri.be/i/lXXv5Q

### DIFF
--- a/wp-content/themes/core/assets/pcss/global/reset.pcss
+++ b/wp-content/themes/core/assets/pcss/global/reset.pcss
@@ -69,7 +69,7 @@ picture {
  * currently be breaking, maybe something to keep an eye on.
  * @see https://github.com/wpcomvip/careforth/pull/14
  */
-iframe
+iframe,
 video,
 embed {
 	max-width: 100%;
--- a/wp-content/plugins/core/src/Blocks/Blocks_Definer.php
+++ b/wp-content/plugins/core/src/Blocks/Blocks_Definer.php
@@ -5,6 +5,7 @@ namespace Tribe\Plugin\Blocks;
 use DI;
 use Tribe\Libs\Container\Definer_Interface;
 use Tribe\Plugin\Blocks\Filters\Contracts\Filter_Factory;
+use Tribe\Plugin\Blocks\Filters\Post_Terms_Filter;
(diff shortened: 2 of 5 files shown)


wp-content/themes/core/assets/pcss/global/reset.pcss
```diff
@@ -63,6 +63,12 @@ picture {
 	max-width: 100%;
 }
 
+/**
+ * @todo Need to ensure we can properly enqueue this reset to prevent pattern
+ * previews from disappearing in the editor. Although, they don't appear to
+ * currently be breaking, maybe something to keep an eye on.
+ * @see https://github.com/wpcomvip/careforth/pull/14
+ */
 iframe
```
ChrisMKindred, line 72: I know this isn't part of your edits, but are we missing a comma after iframe?  All the other CSS properties are listed single line with a comma after, and this one isn't...

ChrisMKindred (approved):

GeoffDusome:
Closing in favor of PR https://github.com/moderntribe/moose/pull/81. Much more the direction I wanted to go with this, I just wanted to see if something like this could be done.

PR #67, opened 2023-07-26
[TRY] Block Animations

GeoffDusome:
## What does this do/fix?

- Adds editor code to create animation related settings for all core blocks.
- Settings include Style (fade, fade in, fade out, etc), Speed, Delay, Easing and two options specific to when/how animations trigger.
- Settings can be overridden by values set within `theme.json`.
- Blocks can be includes or excluded from having animation settings (likely would never need to use both of these settings, probably exclude would be more worthwhile).
- Adds JS to add classes to blocks that should animate via `IntersectionObserver`.
- Adds styles to actually trigger the animations
- Please note that the settings are just arbitrary values chosen by me and can absolutely be updated to be anything, we're not locked down to anything here. 
- Closes #52 in favor of this ticket, since that code is included in this PR.

## QA

Screenshots/video:
- Demo Video: https://www.loom.com/share/13e7c18144f143f8b05960ecd5f135d7


wp-content/themes/core/assets/js/theme/ready.js
```diff
@@ -33,6 +34,10 @@ const init = () => {
 
 	bindEvents();
 
+	// initialize on scroll animations
+
+	animateOnScroll();
```
LayaTaal, line 39: :bulb: might be nice to have a global theme setting so this can be optionally turned off for sites that don't want it.

wp-content/themes/core/assets/js/admin/block-animation.js
```diff
@@ -0,0 +1,354 @@
...
+					{ isSelected && (
+						<InspectorControls>
+							<PanelBody
+								title="Tribe Animations"
+								initialOpen={ false }
+							>
+								<SelectControl
+									label="Animation Style"
+									value={ animationStyle ?? 'none' }
+									help="Animation style is the type of animation you want to display."
```
LayaTaal, line 116: :question: Do we need to internationalize all of these strings?

wp-content/themes/core/assets/js/admin/block-animation.js
```diff
@@ -0,0 +1,354 @@
...
+									help="Animation speed is the speed at which the animation should run."
+									onChange={ ( newValue ) =>
+										setAttributes( {
+											animationSpeed: newValue,
+										} )
+									}
+									options={ [
+										{
+											label: __( 'Extra Slow', 'tribe' ),
+											value: '0.7s',
```
LayaTaal, line 170: Having the animation values hard-coded in here feels odd to me. Could we give control over to the admin user to define an array of animation speeds? The same applies to delays, and easing as well.

Alternatively, if we want to keep this under the hood, maybe we could start a constants file to contain all of this kind of knowledge in one location. That might make it easier to maintain down the road.

wp-content/themes/core/assets/js/admin/block-animation.js
```diff
@@ -0,0 +1,354 @@
+import { InspectorControls } from '@wordpress/block-editor';
```
LayaTaal, line 1: :taco: @GeoffDusome awesome work setting up the admin controls!

LayaTaal (commented):
@GeoffDusome I know I'm not tagged here but I just wanted to leave some praise and start a discussion on a few questions I have.

First of all, awesome job going through this and making a really clean implementation that works for core blocks. I love it and I absolutely don't want to detract from all of the hard work you did here. 

I have a few larger questions and then some more specific comments in the code.

1. Instead of rolling our own animation functions, why not interface with an existing, more robust animation library? Could we get more bang for the buck with that approach?
2. It feels like this should all be moved to a plugin instead of housed within our theme. If we make updates, sites using older implementations of Moose won't get the advantages of those updates. Separating this into a plugin would also make maintaining this code easier.
3. Speaking of plugins, have we tried using any existing plugins out there to add animations to blocks?

wp-content/themes/core/assets/js/admin/block-animation.js
```diff
@@ -0,0 +1,413 @@
...
+			style: blockStyles,
+		};
+
+		return (
+			<Fragment>
+				<div style={ blockStyles } className={ blockClass }>
+					<BlockEdit { ...blockProps } />
+					{ isSelected && (
+						<InspectorControls>
+							<PanelBody
```
LayaTaal, line 184: @GeoffDusome it might be nice to start our own `/components` directory and pull this whole piece out into its own component. That way if we are creating custom blocks in the future we can just do something like below and not have to add/edit this everywhere.

```
<InspectorControls>
    <AnimationControls />
</ InspectorControls>

GeoffDusome:
Closing in favor of implementing block animations into Moose Resources: https://github.com/moderntribe/moose-resources/pull/2

PR #73, opened 2023-07-28
[FIX] update block registration function

GeoffDusome:
## What does this do/fix?

- Use `register_block_type_from_metadata` function (instead of `register_block_type`, which relies on the "name") for registering blocks since we're registering via the `block.json` file. `register_block_type` should work, but it makes a bit more sense to use the function that specifically is looking for the `block.json` file.


ChrisMKindred (approved):
🌮

PR #74, opened 2023-07-28
[MOOSE-36] Search Results Template

GeoffDusome:
## What does this do/fix?

- finalizes search results template that was started in the post card pattern PRs.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-36)

Screenshots/video:
image

### DIFF
--- a/wp-content/themes/core/templates/search.html
+++ b/wp-content/themes/core/templates/search.html
@@ -10,7 +10,7 @@
 
 <!-- wp:tribe/query-results-count {"align":"wide","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|30"}}}} /-->
 
-<!-- wp:query {"queryId":0,"query":{"perPage":1,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true,"taxQuery":null,"parents":[]},"displayLayout":{"type":"list"},"align":"wide"} -->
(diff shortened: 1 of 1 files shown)


wp-content/themes/core/templates/search.html
```diff
@@ -10,7 +10,7 @@ <h1 class="wp-block-heading has-text-align-center" style="margin-top:0">Search</
 
 <!-- wp:tribe/query-results-count {"align":"wide","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|30"}}}} /-->
 
-<!-- wp:query {"queryId":0,"query":{"perPage":1,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true,"taxQuery":null,"parents":[]},"displayLayout":{"type":"list"},"align":"wide"} -->
+<!-- wp:query {"queryId":0,"query":{"perPage":"16","pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":true,"taxQuery":null,"parents":[]},"displayLayout":{"type":"list","columns":3},"align":"wide"} -->
```
dpellenwood, line 13: Can we leave this blank/default to the wp-admin setting for post-per-page?

dpellenwood (approved):
One suggestion, but this looks good, @GeoffDusome Thanks for seeing it through!

wp-content/themes/core/templates/search.html
GeoffDusome: I believe the `inherit` attributes overrides all the other attributes set, but all of the attributes are necessary for the block for some reason. I did a test, changing the global setting for posts per page and confirmed that this query is pulling via that global setting and not the local block settings. It's quite confusing to someone reading the pattern for sure.

PR #75, opened 2023-07-31
[TRY] Column Block Stacking Order

GeoffDusome:
## What does this do/fix?

- adds "Stacking Order" controls to the Column block (not Columns block, each individual Column block)
- this will allow designers to change the order of columns in a Columns block to match the layout they want for mobile. 
- "Mobile" is set to trigger at the "small max" media width we have defined (`781px`), which is when the Columns block also breaks to mobile.
- setting a stacking order sets a class on the column as well as a CSS custom property. Which then feeds into a selector in our FE CSS that determines when the custom properly should be used and what value it should have.
- this uses an experimental `NumberControl` component, which could change at some point, leaving some risk for this item specifically.
- pretty much uses the same logic as we used for the block animations code, which made implementing this a lot simpler. This maybe took an hour and a half to implement versus the block animation stuff where I was still learning which was probably closer to 15 hours total

## QA

Screenshots/video:
- Quick demo video: http://p.tri.be/v/V31pJg
- Longer demo video: (coming soon)


GeoffDusome:
Feel free to review this at any point but I'll put together a quick demo video in the morning. I have this as dependent on the block animation stuff, but it likely doesn't need to be.

wp-content/themes/core/assets/js/admin/stacking-order.js
```diff
@@ -0,0 +1,181 @@
...
+		'tribe/stacking-order-advanced-control',
+		stackingOrderControls
+	);
+
+	addFilter(
+		'blocks.getSaveContent.extraProps',
+		'tribe/apply-stacking-order-classes',
+		applyStackingOrderProps
+	);
+};
```
ChrisMKindred, line 169: This seems overly repetitive calling multiple consts with `init` and `registerFilters`. Can't we just include the `addFilter()` functions in the `init` const and break it out in the future in we need to.

wp-content/themes/core/assets/pcss/global/stacking-order.pcss
```diff
@@ -0,0 +1,14 @@
+:root {
+	--tribe-stacking-order: 0;
+}
+
+/**
+ * need to target .is-not-stacked-on-mobile to make sure
+ * stacking order actually makes sense
+ */
+.wp-block-columns:not(.is-not-stacked-on-mobile) .wp-block-column.tribe-has-stacking-order {
```
ChrisMKindred, line 9: I love that we aren't neutering the `.is-not-stacked-on-mobile` option on the columns block. 🌮

ChrisMKindred (commented):
Great work @GeoffDusome.  

I have added a couple of comments I want to make sure I can find when talking to the practice team about implementing best practices for this kind of thing around naming conventions. I want to ensure everybody is on the same page before giving the full go-ahead to merge, but this is exactly the direction we need to go.

wp-content/themes/core/assets/js/admin/stacking-order.js
```diff
@@ -0,0 +1,181 @@
...
+					{ isSelected && (
+						<InspectorControls>
+							<PanelBody
+								title={ __( 'Tribe Stacking Order', 'tribe' ) }
+								initialOpen={ false }
+							>
+								<NumberControl
+									label={ __( 'Stacking Order', 'tribe' ) }
+									value={ stackingOrder ?? 0 }
+									help={ __(
```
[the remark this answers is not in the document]
GeoffDusome: I was talking with @jamiepaul as I was putting this together and I had asked if we should put an "on/off" toggle within these controls. I could look at if we have access to that property, wouldn't be very hard to throw a `hidden` check on the `NumberControl` depending on that property existing/being turned on.

wp-content/themes/core/assets/js/admin/stacking-order.js
```diff
@@ -0,0 +1,181 @@
...
+		};
+
+		return (
+			<Fragment>
+				<div style={ blockStyles } className={ blockClass }>
+					<BlockEdit { ...blockProps } />
+					{ isSelected && (
+						<InspectorControls>
+							<PanelBody
+								title={ __( 'Tribe Stacking Order', 'tribe' ) }
```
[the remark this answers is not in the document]
GeoffDusome: This makes sense, I'll update.

wp-content/themes/core/assets/js/admin/stacking-order.js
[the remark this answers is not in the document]
ChrisMKindred: Yeah, I did a bunch of research on that when I started looking at the PR. I would mark it up as a nice to have in the future.

wp-content/themes/core/assets/js/admin/stacking-order.js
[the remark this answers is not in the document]
dpellenwood: Ha. Was going to say the exact same thing (after the demo). Looking at the code it's already cleaned up. Love it!

wp-content/themes/core/assets/js/admin/stacking-order.js
```diff
@@ -0,0 +1,164 @@
+/**
+ * @module stacking-order
```
dpellenwood, line 2: This will end up in an editor (rather than admin) enqueue with your other PR, correct?

wp-content/themes/core/assets/js/admin/ready.js
```diff
@@ -49,6 +50,9 @@ const domReady = () => {
 	// initialize admin block animation settings
 	animation();
 
+	// initialize admin stacking order settings
+	stackingOrder();
```
dpellenwood, line 54: I wonder if we can look into a core/column block name space/storage location for this script. I don't like adding it as a "global" or other generally-located script file in the directory structure. (applicable to both the script and the pcss partial.)

dpellenwood (approved):
@GeoffDusome , I love this. Such a great improvement to the editor experience. I'd like to see a couple changes to how where the pcss/js partials are stored and how they are enqueued, but otherwise, this is good to go!

wp-content/themes/core/assets/js/admin/ready.js
GeoffDusome: I've moved the code to the `column` directory under `blocks/core/`. This code should now be attached to the Column block, rather than attached globally. The name is currently `admin.js` for that file, but this will be updated to `editor.js` with the changes to enqueuing.

PR #76, opened 2023-08-03
[MOOSE-32] Post Single Template

GeoffDusome:
## What does this do/fix?

- adds post author name block support & styles
- adds "Tag" text style
- targets wrapper on the main post content in order to add paragraph spacing. We'll possibly likely want to target some other blocks that are used often within post content to add spacing. This point is heavily dependent on the project. By default Moose doesn't add any spacing to any core blocks. But if we add default spacing to core blocks in `theme.json` then this code won't be necessary. Maybe we should figure this out before moving further. 
- styles tags in the Post Terms block by adding a `.is-tags` class. This is currently the only way to style this block differently. Whether or not we create separate blocks to handle different instances of this type of content, we will always need to target either of these in this block as the same block is used for categories & tags lists. 
- moves both `page.html` and `single.html` to use very simple pattern with only the header, footer, and post content. This allows us the most control.
- adds code to automatically pull the new "Post" pattern into the template. This allows any aspect of the page to be edited to the liking of the client. So while we're offering a starting place, the client doesn't need to stick to that one template and one place to add content. 
- **[TRY]** filters post content block to make the wrapper a `main` element instead of a `div`

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-32)

Screenshots/video:
image


wp-content/themes/core/templates/page.html
```diff
@@ -1,5 +1,3 @@
 <!-- wp:template-part {"slug":"header","tagName":"header","className":"site-header"} /-->
-<!-- wp:group {"tagName":"main","layout":{"type":"constrained"}} -->
-<main class="wp-block-group"><!-- wp:post-content {"align":"full","layout":{"type":"constrained"}} /--></main>
-<!-- /wp:group -->
+<!-- wp:post-content {"align":"full","layout":{"type":"constrained"}} /-->
```
jamiepaul, line 2: 🔴 we should keep the `main` element on the template. Is it possible to change the post-content block's tag to `main`?

wp-content/themes/core/templates/single.html
```diff
@@ -1,69 +1,3 @@
...
-	<!-- wp:comments-pagination-next /-->
-	<!-- /wp:comments-pagination -->
-	</div>
-	<!-- /wp:comments-query-loop -->
-	<!-- wp:post-comments-form /-->
-</main>
-<!-- /wp:group -->
-<!-- wp:template-part {"slug":"footer","tagName":"footer","className":"site-footer"} /-->
\ No newline at end of file
+<!-- wp:post-content {"align":"full","layout":{"type":"constrained"}} /-->
```
jamiepaul, line 2: 🔴 same here, we don't want to lose the `main` element

wp-content/themes/core/blocks/core/postterms/style.pcss
```diff
@@ -2,10 +2,50 @@
...
+
+	/**
+	 * Adding the .is-tags class to the post terms block will give pill shaped
+	 * clickable tags: http://p.tri.be/i/K0SeDF
+	 * Post terms block adds forced separator with an empty space if nothing is
+	 * defined in the "Separator" field in the block editor, so the gap of 2px
+	 * is set to define a small space between each tag and the separator (for
+	 * a total of ~8px between each tag)
+	 */
+	&.is-tags {
```
jamiepaul, line 27: 🟠  looks like `taxonomy-post_tag` is added to the block. Was there a reason to use a custom class vs targeting what comes from WP? I'd prefer if we targeted that vs devs/editors knowing that class controls something

wp-content/themes/core/blocks/core/postauthorname/style.pcss
```diff
@@ -0,0 +1,25 @@
...
+	@mixin t-body-small;
+
+	/**
+	 * add ::before pseduo element to create "By" text
+	 */
+	&::before {
+		content: "By ";
+	}
+
+	> a {
```
jamiepaul, line 16: Lets target the class `wp-block-post-author-name__link` here vs the tag for lower specificity

jamiepaul (changes requested):
🔴 couple things to touch up here but overall this is great work @GeoffDusome it'll be nice to have this built into Moose

wp-content/themes/core/templates/single.html
jamiepaul:
```suggestion
<!-- wp:group {"tagName":"main"} -->
<main class="wp-block-group">
	<!-- wp:post-content {"align":"full","layout":{"type":"constrained"}} /-->
</main>
<!-- /wp:group -->
```

@GeoffDusome maybe I'm missing something but I just tested this in my browser- if we don't filter the post content block to make it a `main` tag, and _instead_ just wrap the post content block like this the layout, padding, etc all looks the same.

The key is that the `main` group block wrapper can't have layout constrained. `.has-global-padding` only gets added to constrained blocks. Since we're constraining the post content block, we don't need `main` to be constrained and can therefore avoid a bunch of nested `.has-global-padding` containers causing layout issues.

Filtering seems like an over-engineered solution if we can just make the update in the template file. Let me know what you think.

GeoffDusome: Agreed. Updated to match the above for both Post & Page.

jamiepaul (approved):
🟢 🙌🏻  this is great @GeoffDusome love that we applied it as the default blocks on post creation

PR #79, opened 2023-08-03
[MOOSE-34] Post Archive Templates

GeoffDusome:
## What does this do/fix?

- adds template markup for `archive.html` & `index.html`. The default posts template runs off of `index.html` currently but I'm wondering if we should override that and have `index.html` be pretty general.
- included updated composer file for RC3 but this is likely to change before this PR gets merged since 6.3 comes out soon
- updates mobile breakpoint for post cards
- styles post template block including additional styles for 2-4 columns. **Note: query blocks have changed their layout to grid layout instead of flex, just something to keep in mind.**

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-34)

Screenshots/video:
image


composer.json
```diff
@@ -25,7 +25,7 @@
 			"@php -r \"file_exists('local-config.php') || copy('local-config-sample.php', 'local-config.php');\"",
 			"@php -r \"file_exists('local-config.json') || copy('local-config-sample.json', 'local-config.json');\""
 		],
-		"setup-wordpress": "./vendor/bin/wp core download --version=6.2.2 --quiet --skip-content --force",
+		"setup-wordpress": "./vendor/bin/wp core download --version=6.3-RC3 --quiet --skip-content --force",
```
dpellenwood, line 28: Good call testing the 6.3 RCs. Let's hold on this until 6.3 comes out (today?!). and update to that as part of a chore branch.

wp-content/themes/core/blocks/core/posttemplate/style.pcss
```diff
@@ -0,0 +1,50 @@
+/**
+ * Styles specific to this block
+ */
+
+/**
+ * Vertical (list layout) post template spacing
+ */
+.wp-block-post-template:not(.is-layout-grid) .wp-block-post + .wp-block-post {
+	margin-top: var(--spacer-40);
```
dpellenwood, line 9: Can we use `gap` preferably or `margin-block-start`?

wp-content/themes/core/blocks/core/posttemplate/style.pcss
```diff
@@ -0,0 +1,50 @@
...
+		grid-template-columns: 1fr;
+
+		@media (--mq-wp-medium) {
+			grid-template-columns: repeat(2, minmax(0, 1fr));
+		}
+
+		@media (--viewport-full) {
+			grid-template-columns: repeat(4, minmax(0, 1fr));
+		}
+	}
```
dpellenwood, line 49: A couple things:
* `grid-template-columns: 1fr;` could be defined within the top-most selector since it's applicable to all the column-count conditions.
* Rather than repeating the actual style definition for `grid-template-columns`, I'd suggest setting a css custom prop for the number of columns that is updated within the class and/or viewport size/media query blocks? It could  default to 2, reducing this to only the conditionals for 3 columns at full and 4 columns at full. Something like `grid-template-columns: repeat(var(--post-template-column-count), minmax(0, 1fr));`

wp-content/themes/core/templates/archive.html
```diff
@@ -1,21 +1,31 @@
...
-		<!-- wp:post-date /-->
-		<!-- wp:post-terms {"term":"category"} /-->
-		<!-- wp:post-excerpt /-->
-	<!-- /wp:post-template -->
-	<!-- wp:template-part {"slug":"pagination"} /-->
-	</div>
-	<!-- /wp:query -->
-</main>
+<!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"0","bottom":"0"},"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|50"}}},"layout":{"type":"constrained"}} -->
+<main class="wp-block-group" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--50)"><!-- wp:query-title {"type":"archive","textAlign":"center","showPrefix":false,"align":"wide","style":{"spacing":{"margin":{"top":"0"}}}} /-->
```
dpellenwood, line 3: Throughout: I'd like to see us start using margin/padding block/inline in place of top/bottom & left/right.

dpellenwood (changes requested):
This is looking pretty good, @GeoffDusome . A couple suggestions & changes, if you'd be so kind.

wp-content/themes/core/templates/archive.html
GeoffDusome: I'm not sure I follow. These are the parameters that get output from WordPress. I don't think we have control over the CSS properties used, right?

wp-content/themes/core/blocks/core/posttemplate/style.pcss
GeoffDusome: `gap` isn't available in this context. In 6.3 when the post template block isn't set to be a grid, it doesn't get any sort of flex layout at all. Just plain block layout. I've updated to `margin-block-start`, though.

wp-content/themes/core/blocks/core/posttemplate/style.pcss
GeoffDusome: I've updated this. I had to add an extra breakpoint to the 3 columns layout though, because it needs to be one column at medium width instead of 2.

composer.json
GeoffDusome: I've reverted this to 6.2.2 in this branch. We'll update to 6.3 separately.

wp-content/themes/core/templates/archive.html
dpellenwood: Ah, duh. Ignore me.

wp-content/themes/core/blocks/core/posttemplate/style.pcss
```diff
@@ -0,0 +1,46 @@
...
+
+	&.columns-4 {
+
+		@media (--viewport-full) {
+			--post-template-column-count: 4;
+		}
+	}
+}
```
dpellenwood, line 46: Just spitballing: I wonder if this is any more readible/better?

```suggestion
.wp-block-post-template.is-layout-grid {
	--post-template-grid-template-columns: 1fr;
	gap: var(--spacer-40);
	grid-template-columns: var(--post-template-grid-template-columns);
	
	&.columns-2,
	&.columns-4 {

		@media (--mq-wp-medium) {
			--post-template-grid-template-columns: repeat(2, minmax(0, 1fr));
		}
	}

	&.columns-3 {

		@media (--viewport-full) {
			--post-template-grid-template-columns: repeat(3, minmax(0, 1fr));
		}
	}

	&.columns-4 {

		@media (--viewport-full) {
			--post-template-grid-template-columns: repeat(4, minmax(0, 1fr));
		}
	}	
}
```

dpellenwood (approved):
Looks great! One possible change to the grid-columns definition. (I'm still thinking there's a better way...)

PR #80, opened 2023-08-04
FE layout fix for content width blocks nested in full width blocks

jamiepaul:
## What does this do/fix?

WP changed some styling when `useRootPaddingAwareAlignments` = true in `theme.json`

Due to the change, blocks w/ align none (settings.layout.contentSize) that are nested inside `.alignfull` blocks were bumping up to the viewport edge on small screens.

This is a common scenario in patterns we've built on projects thus far, where we use a group block aligned full to wrap a pattern. Inside the pattern, heading and paragraph blocks are aligned none. 

Notes:
- the `settings.layout.wideSize` max-width value is a calculation that takes into account the grid margins, so this fix is not necessary
- Because we're targeting the inner blocks with `:not(.alignfull):not(.alignwide)` the fix will apply to any additional width settings added for the Block Editor Custom Alignments plugin

Screenshots/video:
- [Link to Video](https://www.loom.com/share/7756959962ac4e33a9449fdd14614b83?sid=4c899302-38d3-49e1-b941-a8cad890346f)


GeoffDusome (approved):
🟢 This looks good and makes sense. Tried it out on a few things and seems to work correctly. I guess if you wanted to fix this as an editor you would just wrap the inner blocks in another group and align that not full. Not an ideal fix, but I think WP's implementation of this kinda makes sense.

dpellenwood (approved):
Thanks for grabbing this, @jamiepaul . Looks good!

PR #81, opened 2023-08-04
[MOOSE-52] Terms Block

GeoffDusome:
## What does this do/fix?

- Adds Terms block
- The Terms block will attempt to replace the Categories and Tags blocks we've been using in the past couple projects. These other blocks have been less than ideal for implementation, so I decided it was time for a new approach.
- With my learnings about core/custom blocks recently I was able to implement attributes, and settings in order to be able to change server-side rendered markup on the FE.
- This block has the same implementation as the other 3 blocks we've been working on, with the addition of attributes.
- The block itself has some simple settings (http://p.tri.be/i/6Qvnk5) that allow it to grab the necessary information it needs (through attributes) in order to render the proper markup for displaying the terms.
- A few notes on the attributes:
    - The `taxonomyToUse` attribute is currently pulling all available taxonomies. I think I could maybe edit this to pull only taxonomies that are attached to the current post type?
    - The `onlyPrimaryTerm` attribute determines if we should grab the full list of terms or only the "primary" term. If Yoast is enabled on the site and we're using the "Category" taxonomy, the code will look for the primary category as designated by Yoast, with a fallback to the first category in the list. If Yoast is not enabled, we just grab the first term in the list returned.
    - The `hasLinks` attribute determines if the terms should display as links or plain text
    - The `termStyle` attribute determines if the terms should display as basic text or as pill shaped "tags"
- I won't go over the styles here, they are pretty straightforward.
- Please see attached demo video to see it in action, I'd like to get some feedback on this block. I'm not expecting this to be a quick PR, just wanted to get an MVP out there.

For reference, the code that pulls the taxonomies looks like this:

```
const taxonomies = useSelect( ( select ) =>
	select( 'core' ).getTaxonomies( { per_page: -1 } )
);
```

There's not a whole lot of documentation on `useSelect`, so I'm just kinda going by what I'm finding online. I would assume `getTaxonomies` has more attributes than `per_page` but I'll have to look for them.

**This PR closes https://github.com/moderntribe/moose/pull/66**

## QA

Link to issue: https://moderntribe.atlassian.net/browse/MOOSE-52

Screenshots/video:
- Settings: http://p.tri.be/i/6Qvnk5
- Demo Video: https://www.loom.com/share/2a30b7c6e5ac47c88500ab6ed0f7ba5e?sid=f82ddc80-8942-434d-813d-ddcbe16f8176


wp-content/themes/core/blocks/tribe/terms/index.js
```diff
@@ -0,0 +1,66 @@
...
+					d="M189 34.8835C193.418 34.8835 197 38.4652 197 42.8835V93.0196C197 95.1413 196.157 97.1762 194.657 98.6764L99.7925 193.54C96.6683 196.665 91.603 196.665 88.4788 193.54C85.3546 190.416 85.3546 185.351 88.4788 182.227L181 89.7059V42.8835C181 38.4652 184.581 34.8835 189 34.8835Z"
+					fill="currentcolor"
+				/>
+				<path
+					fillRule="evenodd"
+					clipRule="evenodd"
+					d="M119 53.5C119 50.4624 116.537 48 113.5 48C110.462 48 108 50.4624 108 53.5C108 56.5376 110.462 59 113.5 59C116.537 59 119 56.5376 119 53.5ZM113.5 32C125.374 32 135 41.6259 135 53.5C135 65.3741 125.374 75 113.5 75C101.626 75 91.9997 65.3741 91.9997 53.5C91.9997 41.6259 101.626 32 113.5 32Z"
+					fill="currentcolor"
+				/>
+			</svg>
```
ChrisMKindred, line 58: Do we usually compress SVG files when we include them in code?  I know it isn't a huge file but if I pass this through imageOptim the size drops by like 65% (from 2kb to 724 bytes).

wp-content/themes/core/blocks/tribe/terms/render.php
```diff
@@ -0,0 +1,85 @@
...
+				'<li class="%s"><span class="%s">%s</span></li>',
+				esc_attr( "wp-block-tribe-terms__term is-style-$style" ),
+				esc_attr( 'wp-block-tribe-terms__link' ),
+				esc_html( $term->name )
+			);
+		}
+		echo '</ul>';
+	} else {
+		// display if no terms are available
+		echo 'No terms to list.';
```
ChrisMKindred, line 83: Should we output this as an HTML comment?  I don't think we would want to display `No terms to list` since they are not required by every post.

ChrisMKindred (commented):
I have a couple of quick comments but everything looks pretty good.

wp-content/themes/core/blocks/tribe/terms/index.js
GeoffDusome: I'm not sure - This is the first time I'm running things through a tool - ImageOptim doesn't seem to be doing anything with the SVG file I created. Are there any special settings I need to be using? Is there a different tool I can use? Looks like all I can get ImageOptim to do was remove the line breaks - which get automatically added by our JS prettifier anyway - so not super helpful.

wp-content/themes/core/blocks/tribe/terms/render.php
GeoffDusome: Yeah, good call. I think I originally intended this text to only show in the editor, but I think an HTML comment is better anyway.

wp-content/themes/core/blocks/tribe/terms/index.js
ChrisMKindred: The tool doesn't matter as much as the need to compress the SVG before including it. It sounds like compression is something that we need to define better outside of this PR.  👍

ChrisMKindred (approved):

wp-content/themes/core/blocks/tribe/terms/render.php
```diff
@@ -0,0 +1,85 @@
...
+	$terms        = get_the_terms( get_the_ID(), $taxonomy );
+	$primary_term = $terms ? $terms[0] : false;
+} else {
+	$terms = get_the_terms( get_the_ID(), $taxonomy );
+}
+
+if ( ! $terms && ! $primary_term && ! is_admin() ) {
+	return;
+}
+?>
```
dpellenwood, line 38: We have [an existing utility method](https://github.com/moderntribe/moose/blob/main/wp-content/plugins/core/src/Templates/Traits/Primary_Term.php) for determining the determining the primary term. I'd suggest using that.

wp-content/themes/core/blocks/tribe/terms/render.php
```diff
@@ -0,0 +1,85 @@
...
+
+<div <?php echo get_block_wrapper_attributes(); ?>>
```
dpellenwood, line 40: I'd liek to take a stab at cleaning up the PHP a bit here. It's pretty intense. 😬

wp-content/themes/core/blocks/tribe/terms/style.pcss
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/* -----------------------------------------------------------------------
+ * Terms Block - List
+ * ----------------------------------------------------------------------- */
+
+.wp-block-tribe-terms__list {
+	display: inline-flex;
+	align-items: center;
+	gap: 8px;
```
dpellenwood, line 33: SHould we be using a spacer var for this?

wp-content/themes/core/blocks/tribe/terms/style.pcss
```diff
@@ -0,0 +1,82 @@
...
+	justify-content: flex-start;
+
+	&.aligncenter {
+		justify-content: center;
+	}
+
+	&.alignright {
+		justify-content: flex-end;
+	}
+}
```
dpellenwood, line 24: Are there existing classes/styles we could use for this?

wp-content/themes/core/blocks/tribe/terms/style.pcss
```diff
@@ -0,0 +1,82 @@
...
+
+			&:hover,
+			&:focus {
+				background-color: var(--color-black);
+				color: var(--color-white);
+				text-decoration: none;
+			}
+		}
+	}
+}
```
dpellenwood, line 82: Since we can control the markup & classes for this block, I'd love to see us use utility classes for these styles rather than a set of one-off styles specific to this block.

dpellenwood (changes requested):
This is an excellent take on the block. A few changes & suggestions, if you'd be so kind.

wp-content/themes/core/blocks/tribe/terms/render.php
GeoffDusome: I didn't want to make everything a list if it didn't need to be a list, or else I suppose I could remove that whole first block. Is there anything specific that isn't clear about how things display? Would it be helpful to move things out of `sprintf` calls maybe?

wp-content/themes/core/blocks/tribe/terms/style.pcss
GeoffDusome: Not that I know of. `aligncenter` and `alignright` come from the the alignment controls on the block and aren't "custom" to this block.

wp-content/themes/core/blocks/tribe/terms/style.pcss
GeoffDusome: Do you have any ideas on where these styles would live in the global settings? Maybe mixins/utilities in `pcss/cards`? Are you just looking for utilities specific to the pill styling?

wp-content/themes/core/blocks/tribe/terms/style.pcss
GeoffDusome: Updated this to a spacer variable.

GeoffDusome:
Closing in favor of moving the Terms block to Moose Resources: https://github.com/moderntribe/moose-resources/pull/3

Open discussion items here have been moved to TODO list in the new PR.

GeoffDusome:
Reopening to continue iterating within Moose.

wp-content/themes/core/blocks/tribe/terms/render.php
ChrisMKindred: I think this php has been made a little overly complex here.  if you don't mind, @GeoffDusome, I am going to take a  stab at a PR to clean some of this up.

ChrisMKindred (changes requested):
@GeoffDusome, I will put together a quick PR to look into the PHP in the template.

ChrisMKindred (approved):
We should be good from the PHP side now.

wp-content/themes/core/blocks/tribe/terms/index.js
dpellenwood: Created a quick SVG file and ran i through [vecta.io](https://vecta.io/nano). It was able to reduce the file size by more than half. I've attached the before & after. It would be nice to minimize this I agree.
icon-after
icon-before

wp-content/themes/core/blocks/tribe/terms/style.pcss
dpellenwood: I'd say adding them to the typography mixins & helpers like your existing t-tag mixin. The key is creating both the mixin and a related helper/utility class. That way any time we have the ability to add classes ourselves we can just use the utility class, instead of defaulting to the mixin. (Since mixins duplicate the styles, which bloats the overall amount of css being rendered.)

dpellenwood (changes requested):
I followed up on a couple existing conversations. We're _really_ close here, Geoff. Thanks for tackling this!

wp-content/themes/core/blocks/tribe/terms/style.pcss
GeoffDusome: I've updated this so that the term links default to have the `t-category` class, removing that extra CSS. If the terms are set to display as pills, the styles should override with that mixin. Also added utility classes that were missing from some mixins.

wp-content/themes/core/blocks/tribe/terms/index.js
GeoffDusome: I've updated the SVG to the new one provided with one edit. It seemed like I needed the `viewBox` property to work properly, otherwise the icon wouldn't show in context. Other than that, it should be much smaller.

GeoffDusome:
@dpellenwood should be good for another look!

dpellenwood (approved):
:chefskiss: This is excellent. merge away!

PR #82, opened 2023-08-04
[FEATURE] Custom Block Icons

GeoffDusome:
## What does this do/fix?

- worked with design to create some custom icons for our custom blocks we've developed to date. Unfortunately, the only way to add custom icons is to do so when registering the block type in the `index.js` file.
- re-organizes the custom blocks into our "Custom" block category
- the "Terms" block icon will be added in the separate PR

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-54?atlOrigin=eyJpIjoiMDA2NmMxYTRlZDVhNDNmYWIwMTEyNDJkNDJiOGUzNjAiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

Screenshots/video:
- http://p.tri.be/i/yqqpyM


dpellenwood (changes requested):
Thanks for tackling this Geoff.

After speaking with the practice team a bit today we'd like to make a few changes here. Specifically:
* Please categorize these blocks in "Theme" category rather than separately in a "Custom" category. Given that these blocks will generally be used with templating/initial development and aren't designed to be used on  regular basis in post content, it doesn't feel necessary to add a unique section in the block insert for them. (That could change if we find we're creating blocks that are more editor-driven.)
* Please use black and white icons for the blocks to match the existing core icons. We can reach out to design if needed to update the SVGs. We feel it's better if these blocks.icons fit in the selector rather than stand out uniquely.

GeoffDusome:
@dpellenwood Updates have been made, feel free to take another look. "Terms" block icon will be switched separately in it's own PR (which is still open).

dpellenwood (approved):
🐳 💨

PR #83, opened 2023-08-07
[MOOSE-50] Admin Asset Enqueuing

GeoffDusome:
## What does this do/fix?

- moves to using `add_editor_style` for global assets in the admin so our reset doesn't affect anything outside of the editor.
- moves to using very clear language between admin and editor. Each are their own thing and need to be declared as such. Each will have their own styles & scripts. 
    - Editor will use the `enqueue_block_editor_assets` hook to call `add_editor_style` to add styles to the actual editor itself, prefixing all style tags with `.editor-styles-wrapper`. This script is pretty smart, so we don't have to go back and fix style rules we're currently using. But this is worth doing some extensive testing with before using in production, just to make sure all our styles are there. It looked pretty good to me.
    - Admin will continue to use `admin_enqueue_scripts` to call `wp_enqueue_style`. 
    - According to [this WordPress developer guide](https://developer.wordpress.org/block-editor/how-to-guides/javascript/loading-javascript/), you still use `wp_enqueue_script` regardless of the situation 

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-50)

Screenshots/video:
- Pattern previews are now working properly: http://p.tri.be/i/yUjZfm

### DIFF
--- a/wp-content/plugins/core/src/Assets/Assets_Subscriber.php
+++ b/wp-content/plugins/core/src/Assets/Assets_Subscriber.php
@@ -19,6 +19,10 @@ class Assets_Subscriber extends Abstract_Subscriber {
 			$this->container->get( Admin_Assets_Enqueuer::class )->register();
 		}, 10, 0 );
 
+		add_action( 'enqueue_block_editor_assets', function (): void {
+			$this->container->get( Editor_Assets_Enqueuer::class )->register();
+		}, 10, 0 );
+
 		// Uncomment this if you want to enable WP login styles
 		// add_action( 'login_enqueue_scripts', function (): void {
 		// 	$this->container->get( Admin_Assets_Enqueuer::class )->enqueue_login_styles();
--- a/wp-content/themes/core/assets/js/editor/ready.js
+++ b/wp-content/themes/core/assets/js/editor/ready.js
@@ -1,0 +1,29 @@
+/**
+ * @module
+ * @exports ready
+ * @description The core dispatcher for the dom ready event javascript.
+ */
+
+/**
+ * @function init
+ * @description The core dispatcher for init across the codebase.
+ */
+
+const init = () => {
+	// intentionally left blank for now
+
+	console.info(
+		'Moose Editor: Initialized all javascript that targeted document ready.'
+	);
+};
+
+/**
+ * @function ready
+ * @description Export our dom ready enabled init.
+ */
+
+const ready = () => {
+	init();
+};
+
+export default ready;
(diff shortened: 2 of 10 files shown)


GeoffDusome:
I think this is in a good place now. Core block styles are enqueuing how I think they _should_ be and the setup feels a lot better IMO. I still want BE to take a look to see if there's any improvements to be made but it feels a lot better than it did before.

wp-content/plugins/core/src/Blocks/Block_Base.php
```diff
@@ -45,25 +45,26 @@ public function get_block_dependencies(): array {
...
 	 */
-	public function register_block_style(): void {
+	public function register_core_block_variations(): void {
 		if ( ! function_exists( 'register_block_style' ) ) {
 			return;
 		}
 
 		foreach ( $this->get_block_styles() as $name => $label ) {
-			register_block_style( $this->get_block_name(), [
+			\register_block_style( $this->get_block_name(), [
```
dpellenwood, line 57: Curious about the preceding slash?

wp-content/plugins/core/src/Assets/Assets_Subscriber.php
```diff
@@ -19,6 +19,10 @@ public function register(): void {
 			$this->container->get( Admin_Assets_Enqueuer::class )->register();
 		}, 10, 0 );
 
+		add_action( 'enqueue_block_editor_assets', function (): void {
```
dpellenwood, line 22: I think we want to use `enqueue_block_assets` instead based on [this dev note for  Core 6.3](https://make.wordpress.org/core/2023/07/18/miscellaneous-editor-changes-in-wordpress-6-3/#post-editor-iframed).

> It’s also worth noting that enqueue_block_editor_assets should not be used to add stylesheets _for editor content_. Please use enqueue_block_assets instead.

dpellenwood (approved):
One change regarding the editor enqueueing hook. Otherwise, this looks great @GeoffDusome . Thanks for tackling it!

wp-content/themes/core/assets/js/editor/ready.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @function init
+ * @description The core dispatcher for init across the codebase.
+ */
+
+const init = () => {
+	// intentionally left blank for now
+
+	console.info(
+		'Moose Editor: Initialized all javascript that targeted document ready.'
```
ChrisMKindred, line 16: Using `Moose` here doesn't provide any additional value and could be confusing to clients.  Moose is more of an internal thing so we know what we are talking about. We can just have `Editor: ....`

ChrisMKindred (approved):
Looks good!  I have one small comment but nothing that would keep this from being merged.

PR #84, opened 2023-08-09
[CHORE] WP Core / Composer Packages Update

GeoffDusome:
## What does this do/fix?

- WP core to 6.3
- ACF to 6.1.8
- Gravity Forms to 2.7.12
- Local Lando version to 8.1
- Yoast SEO to ^20.1
- Misc composer packages updated to match local PHP version

Note: This update will require a `lando rebuild` as well as a `lando composer install`.


.lando.yml
```diff
@@ -1,7 +1,7 @@
 name: moose
 recipe: wordpress
 config:
-  php: '8.0'
+  php: '8.1'
```
dpellenwood, line 4: @ChrisMKindred thoughts on moving to 8.2 by default? It's supported/recommended by WP Engine, Pantheon, and VIP now. Also, was there a platform setting in composer.json that needs to be adjusted to match this? I seem to recall Justin F mentioning it recently.

dpellenwood (commented):
This looks good, @GeoffDusome . I'd like to hear Chris's input on my questions before we merge though.

dpellenwood:
One additional question @GeoffDusome: did you check that all the other composer-installed plugins (Yoast SEO, etc) are also updated?

.lando.yml
ChrisMKindred: We can't go 8.2 yet. Here is the supported version chart from WP. https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/

8.1 is compatible with 8.2. The project should be changed to 8.2 if the host defaults to it but if we start with 8.2 and the host defaults to 8.1 (or 8.0) then it is possible for us to not be backward compatible.

ChrisMKindred: We can and should add a platform setting for the composer. That can be a different pr though.

GeoffDusome:
@ChrisMKindred I updated the other packages where I felt comfortable, would you mind taking a quick look to ensure we're all up to date?

ChrisMKindred (approved):
Thanks for getting this put together.  I made a small composer update but nothing big.  This should be good to go.

PR #85, opened 2023-08-09
Update Composer Platform and ACF Require

ChrisMKindred:
## What does this do/fix?

Adds a platform option to the composer config to make sure even if you run composer using a different version of php, it will still download for 8.1. 

Changes the composer require for ACF to the [recommended version by WP Engine](https://www.advancedcustomfields.com/resources/installing-acf-pro-with-composer/).

### DIFF
--- a/auth-sample.json
+++ b/auth-sample.json
@@ -2,7 +2,7 @@
 	"http-basic": {
 		"connect.advancedcustomfields.com": {
 			"username": "ACF License Key",
-			"password": "SITE_URL"
+			"password": "https://moose.lando.site"
 		}
 	}
 }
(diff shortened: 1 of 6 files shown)


auth-sample.json
```diff
@@ -0,0 +1,8 @@
+{
+	"http-basic": {
+		"connect.advancedcustomfields.com": {
+			"username": "ACF License Key",
+			"password": "SITE_URL"
+		}
```
dpellenwood, line 6: Will this require the site to be "registered" with the license key ahead of time? If so, we should probably mention that in the docs.

dpellenwood (approved):
One question/comment. Looks good tho!

auth-sample.json
ChrisMKindred: No, there is no registration that is needed.  It will add the site to the active site list in the account section of the ACF site automatically.

GeoffDusome (approved):
🟢 Good to merge after docs update!

PR #86, opened 2023-08-09
[FIX] WP 6.3 Block Styles Support

GeoffDusome:
## What does this do/fix?

- remove block styles support
- update items in styleguide that don't quite match after removing styles


dpellenwood (approved):
Looks great, Geoff. Thanks for tackling this. Do you think there are any styles we've added that we no longer need w/o the core theme styles loading now?

ChrisMKindred (approved):

PR #94, opened 2023-08-16
[FIX] Update Styleguide Pattern for WP 6.3

GeoffDusome:
## What does this do/fix?

- reworks styleguide pattern to be a little bit easier to read & work with, as well as updating all of the blocks to 6.3 so we no longer have any confusion with blocks using old formats.
- move typography for Pullquote & Quote block to a `css` option in `theme.json` - it appears that the `typography` settings for these blocks was being overridden by the Paragraph block. I don't know if it would be better to add the custom styles in our actual block styles or if this would be ok 🤔 

Screenshots/video:
image

### DIFF
--- a/wp-content/themes/core/theme.json
+++ b/wp-content/themes/core/theme.json
@@ -421,11 +421,6 @@
 						"bottom": "var(--wp--preset--spacing--40)"
 					}
 				},
-				"typography": {
-					"fontSize": "var(--wp--preset--font-size--50)",
-					"fontWeight": "var(--wp--custom--font-weight--bold)",
-					"lineHeight": "1.2"
-				},
 				"elements": {
 					"cite": {
 						"typography": {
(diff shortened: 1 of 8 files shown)


wp-content/themes/core/theme.json
```diff
@@ -436,7 +431,8 @@
 							"fontStyle": "normal"
 						}
 					}
-				}
+				},
+				"css": "> blockquote p {font-size: var(--wp--preset--font-size--50); font-weight: var(--wp--custom--font-weight--bold); line-height: 1.2;}"
```
jamiepaul, line 435: 🔴 yea, let's not do this. I know @dpellenwood wasn't a fan and has commented on it before. Let's move the styles to our block's `.pcss` partial

jamiepaul (approved):
🟢 Let's move those styles to the block pcss and this is g2g @GeoffDusome

PR #96, opened 2023-08-24
Asset enqueue refactor

dpellenwood:
## What does this do/fix?
Fixes for asset enqueues with all the different contexts:
* Theme styles being properly loaded on the public website as well as the different editor contexts: post editor (inline or iframe) and site editor.
* Editor-only styles & scripts being properly loaded for both editors & contexts.

A couple things to address yet:
* I think we should go ahead and create a separate reset.pcss for the editor (or just reset styles we need to specifically?) rather than using our public reset.pcss in that context. It's not a blocker right now, but something we could address later.
* The `file_get_contents()` call used for inlining the styles is probably not the exact right approach. I think we need to adjust that for WPVIP's file system, correct?

- [Link to Video](https://www.loom.com/share/44b5c3cbce2a48e48b0e24ea3a5ae959)
- [Inlined block styles reference](https://make.wordpress.org/core/2021/07/01/block-styles-loading-enhancements-in-wordpress-5-8/)

### DIFF
--- a/wp-content/plugins/core/src/Blocks/Block_Base.php
+++ b/wp-content/plugins/core/src/Blocks/Block_Base.php
@@ -26,6 +26,10 @@ abstract class Block_Base {
 		return $this->get_block_name();
 	}
 
+	public function get_block_style_handle(): string {
+		return str_replace( 'core/', 'wp-block-', $this->get_block_name() );
+	}
+
 	/**
 	 * Block styles to be defined in extending class
 	 *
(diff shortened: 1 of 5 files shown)


GeoffDusome:
This is great! Exactly where we needed to get. I'll take a closer look tomorrow morning but I think this will do the trick for all the sites we're needing to update. This update needs to happen for comic-con first so once we're happy with this I'll port it there (should be pretty easy) and then we can deal with QCHI & Careforth afterwards. Thank you so much for taking a look at this, took a huge weight off my shoulders in an already stressful week. 🌮 🌮 🌮 🌮 🌮 

Not sure if this is helpful RE: WPVIP but here's what I found on `file_get_contents()`: https://docs.wpvip.com/technical-references/code-quality-and-best-practices/retrieving-remote-data/

GeoffDusome (approved):
🟢 I'm seeing positive on all the different situations on my end so I'll call this good but I'll wait for input from BE as well in case there's anything that would be blocking in their minds.

wp-content/plugins/core/src/Blocks/Block_Base.php
```diff
@@ -97,17 +83,21 @@ public function enqueue_core_block_editor_front_style(): void {
 			return;
 		}
 
+		wp_add_inline_style( $handle, file_get_contents( $src_path ) );
```
dpellenwood, line 86: @ChrisMKindred or @JMRhodes Just calling this `file_get_contents()` out. I know we had some trouble with WPVIP's file system and want to make sure this is both functional and acceptable from a performance standpoint. This based on [a WP Dev blog post](https://make.wordpress.org/core/2021/07/01/block-styles-loading-enhancements-in-wordpress-5-8/) and is inlining the css for core blocks is the ideal approach here for FE.

JMRhodes: @dpellenwood yes I would stick with `file_get_contents()`. 

VIP just recommends `wpcom_vip_file_get_contents()` as a caching wrapper. But that function is only available in the VIP environments. Even on local development, we don't have access to that function, so it's difficult to test.

ChrisMKindred (approved):

PR #97, opened 2023-08-29
Add accessibility-checker to plugins

ChrisMKindred:
## What does this do/fix?

Adds the Accessibility Checker plugin for checking content as it is entered.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -125,6 +125,7 @@
 		"vinkla/extended-acf": "^13.6",
 		"vlucas/phpdotenv": "^5.5",
 		"wp-cli/wp-cli-bundle": "^2.8",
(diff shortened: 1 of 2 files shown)


dpellenwood:
Thanks for putting this up, Chris. I reviewed the plugin and found a couple issues that I think will prevent us from adopting it. Notably:
* It's not very smart about errors. It’s finding thins like underlined text as a problem, but the underlining is specifically for links, which is preferred.
* It adds a metabox to the editor, which is kind of an anti-pattern now fro WordPress. (settings & meta should be handled in the editor sidebar.)
* Based on your feedback, it makes some strange curl requests to the FE of the site to run the report, which feels hacky.

PR #99, opened 2023-08-31
Replace Yoast SEO with Rank Math SEO

dpellenwood:
## What does this do/fix?

Switches out Yoast SEO for [Rank Math SEO](https://wordpress.org/plugins/seo-by-rank-math/), a bit of a newer SEO plugin for WordPress. In our early evaluations, Rank Math has a better UI, more features, and an importer from Yoast SEO if you want to move an existing project from Yoast to Rank Math.

We'll chat about this at the next Engineering Bi-weekly.


ChrisMKindred (approved):
🚢

GeoffDusome (approved):
🟢

PR #100, opened 2023-09-01
[MOOSE-32] Post Single QA

GeoffDusome:
## What does this do/fix?

- updates paragraphs, headings, to have global spacing
- removes margin definitions from reset
- fixes lists to ensure list styles don't get applied to other blocks
- updates styleguide pattern to remove extra spacers that were used to create space 

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-32)

### DIFF
--- a/wp-content/themes/core/assets/pcss/global/reset.pcss
+++ b/wp-content/themes/core/assets/pcss/global/reset.pcss
@@ -29,13 +29,6 @@ body {
 	overflow-x: hidden;
 }
 
-p,
-ol,
-ul,
-blockquote {
-	margin: 0;
-}
-
 cite,
 figcaption {
 	display: block;
--- a/wp-content/themes/core/blocks/core/lists/style.pcss
+++ b/wp-content/themes/core/blocks/core/lists/style.pcss
@@ -6,7 +6,7 @@
  * Ordered List
  * ---------------------------------------------------------------------- */
 
-:where(.wp-site-blocks) ol,
+.wp-site-blocks ol:not([class]),
 ol.wp-block-list {
 	padding-inline-start: 1.2rem;
 
@@ -27,7 +27,7 @@ ol.wp-block-list {
  * Unordered List
  * ---------------------------------------------------------------------- */
 
-:where(.wp-site-blocks) ul,
+.wp-site-blocks ul:not([class]),
 ul.wp-block-list {
 	padding-inline-start: 10px;
 
--- a/wp-content/themes/core/theme.json
+++ b/wp-content/themes/core/theme.json
@@ -372,6 +372,12 @@
 		},
 		"blocks": {
 			"core/paragraph": {
+				"spacing": {
+					"margin": {
+						"top": "1em",
+						"bottom": "0"
+					}
+				},
 				"typography": {
 					"fontSize": "var(--wp--preset--font-size--20)",
 					"fontWeight": "var(--wp--custom--font-weight--regular)",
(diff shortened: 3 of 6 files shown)


jamiepaul (approved):
🟢 this looks good to me @GeoffDusome

wp-content/themes/core/assets/pcss/global/reset.pcss
```diff
@@ -29,13 +29,6 @@ body {
 	overflow-x: hidden;
 }
 
-p,
-ol,
-ul,
-blockquote {
```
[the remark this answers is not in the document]
GeoffDusome: For paragraphs and lists specifically, this PR adds vertical margins to these elements. 
The core blocks that use `blockquote` now should have their own rules for handling inner paragraphs. For the `pullquote` block, I've removed the default margin around the inner `blockquote` element.

wp-content/themes/core/theme.json
```diff
@@ -372,6 +372,12 @@
 		},
 		"blocks": {
 			"core/paragraph": {
+				"spacing": {
+					"margin": {
```
dpellenwood, line 376: Before we implement this, I'd prefer to get an updated Figma comp from design with the spacing redlined. There's an outstanding question to the design team in the [Figma file already](https://www.figma.com/file/zbjC33tZJz5A901jeXM7zv/Website-Design-System?type=design&node-id=1975-3703&mode=design&t=rhAkHHysAFWRe6Gj-0). We can track there.

dpellenwood (commented):
This is good with me. But, I'd like to wait until we get design feedback on the spacing values (per my comment).Thanks Geoff!

wp-content/themes/core/theme.json
GeoffDusome: Core block spacing has been updated to match Figma.

wp-content/themes/core/blocks/core/lists/style.pcss
```diff
@@ -6,7 +6,7 @@
  * Ordered List
  * ---------------------------------------------------------------------- */
 
-:where(.wp-site-blocks) ol,
+.wp-site-blocks ol:not([class]),
```
dpellenwood, line 9: Talk me through this selector & change:

It looks like we're targeting an `ol` inside `.wp-site-blocks` that does not have a class attribute, correct? Then, on the next line, we're also targeting any `ol` _with_ a class of `wp-block-list`? 

I'm guessing this is needed for differences between the editor and the FE, but I'm wondering if we can can reduce/simplify it more?

dpellenwood (approved):
Changes look good, @GeoffDusome . I do have one question about the list selectors. Thanks for tackling this.

wp-content/themes/core/blocks/core/lists/style.pcss
GeoffDusome: The styles in this file are supposed to be specific to the List block. The problem is that the list block (much like the paragraph block) doesn't get a class on the FE, which makes it difficult to style and we've been having these styles bleed out into other blocks. Fortunately, other blocks that use lists on the FE _do_ get classes (Post Template block / Query block was a big pain point, Categories block, etc.). I've found the only list elements that don't get classes come from the list block (ironic 🙄).

In the editor, the List block _does_ get a class of `wp-block-list`, so that's easy to target and we don't need to prefix, because the `editor-styles-wrapper` prefix happens automatically. 

The FE selector will also get prefixed but it won't do anything because I believe `editor-styles-wrapper` and `wp-site-blocks` are the same element in the editor.

dpellenwood: Thanks for clarifying. Merge away!

PR #102, opened 2023-09-12
[MOOSE-34] Misc Template Fixes

GeoffDusome:
## What does this do/fix?

- Adds "Query No Results" block support so that we can add a "hack" to remove the no results block on the search template where we actually have results but it's the last page (thanks for finding this one @jamiepaul)
- Moves post card responsive to container queries, which should make the responsive for these cards feel a lot better.
- Adds explanations to the list block selectors
- The post template block was affected negatively by the asset enqueuing update. The selectors weren't specific enough to override some WP styling. I've added a `.wp-site-blocks` prefix to these selectors. @dpellenwood, I was incorrect on the list note from the other PR. `wp-site-blocks` is a child element of `editor-styles-wrapper`, so we _are_ able to also use that as a selector in the editor. The list block styles weren't affected by this change, however, as those selectors still work correctly in the editor & on the FE.
- Added some style margin to the top of the pagination block - margin settings are not available on this block.
- Since we added global margins to paragraphs, there were a few places where paragraphs were now getting spacing where they shouldn't. I've removed this extra spacing from templates & style files where it seemed reasonable to do so.
- Both archive templates (`archive.html` & `index.html`) lost the bottom margin on their headlines. I've added this back into the templates.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-34)

### DIFF
--- a/wp-content/themes/core/assets/pcss/cards/post.pcss
+++ b/wp-content/themes/core/assets/pcss/cards/post.pcss
@@ -14,6 +14,7 @@
 
 .p-card-post {
 	position: relative;
+	container: postcard / inline-size;
 }
 
 /* -----------------------------------------------------------------------
@@ -34,7 +35,7 @@
 		}
 	}
 
-	@media (--viewport-small-max) {
+	@container (max-width: 320px) {
 		width: 22% !important;
 		float: right;
 		margin-inline-start: 2em;
--- a/wp-content/themes/core/blocks/core/lists/style.pcss
+++ b/wp-content/themes/core/blocks/core/lists/style.pcss
@@ -4,6 +4,12 @@
 
 /* ----------------------------------------------------------------------
  * Ordered List
+ *
+ * The first selector applies to the FE, the second in the editor.
+ * The list block is the only block on the FE that doesn't get a class,
+ * so we're able to apply a :not() selector for the class attribute.
+ * While both will apply in the editor, only the second will affect
+ * the actual list block.
  * ---------------------------------------------------------------------- */
 
 .wp-site-blocks ol:not([class]),
@@ -25,6 +31,12 @@ ol.wp-block-list {
 
 /* ----------------------------------------------------------------------
  * Unordered List
+ *
+ * The first selector applies to the FE, the second in the editor.
+ * The list block is the only block on the FE that doesn't get a class,
+ * so we're able to apply a :not() selector for the class attribute.
+ * While both will apply in the editor, only the second will affect
+ * the actual list block.
  * ---------------------------------------------------------------------- */
 
 .wp-site-blocks ul:not([class]),
--- a/wp-content/themes/core/blocks/core/querynoresults/style.pcss
+++ b/wp-content/themes/core/blocks/core/querynoresults/style.pcss
@@ -7,6 +7,7 @@
 	/**
 	* Hack to get around query block displaying no results blocks
 	* on final page of results
+	* @see https://github.com/WordPress/gutenberg/pull/53772
 	*/
 	.search-results & {
 		display: none;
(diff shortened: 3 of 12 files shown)


wp-content/themes/core/blocks/tribe/query-results-count/style.pcss
```diff
@@ -9,6 +9,10 @@
 
 	@mixin t-body;
 
+	> p {
```
jamiepaul, line 12: 🟠 It looks like the paragraph block in the Query No Results already has the margin top set to zero, so maybe we don't need this as a custom style?

If we do need it as a custom style, I'd rather see this applied w/ a custom class on the specific paragraph. It's possible some projects will have multiple paragraph blocks and we don't want to affect others

jamiepaul (approved):
This looks good except for one custom style that hopefully just isn't needed. Thanks for porting over the no results fix @GeoffDusome 🙌🏻

wp-content/themes/core/blocks/tribe/query-results-count/style.pcss
GeoffDusome: This is the query results count block, which is different than the no results block. This block only has the one paragraph in it and should never have more than one. With every file named the same it gets confusing 😂

wp-content/themes/core/assets/pcss/cards/post.pcss
```diff
@@ -34,7 +35,7 @@
 		}
 	}
 
-	@media (--viewport-small-max) {
+	@container (max-width: 320px) {
```
dpellenwood, line 38: 😍

wp-content/themes/core/blocks/core/lists/style.pcss
```diff
@@ -4,6 +4,12 @@
...
  * Ordered List
+ *
+ * The first selector applies to the FE, the second in the editor.
+ * The list block is the only block on the FE that doesn't get a class,
+ * so we're able to apply a :not() selector for the class attribute.
+ * While both will apply in the editor, only the second will affect
+ * the actual list block.
  * ---------------------------------------------------------------------- */
 
 .wp-site-blocks ol:not([class]),
```
dpellenwood, line 15: ❓  Just so I'm clear, the selector `.wp-site-blocks` IS rendered on the front-end somewhere? Is that a class we're adding to the templates somewhere? I'm not familiar with that class in the output of a public site.

wp-content/themes/core/blocks/core/querynoresults/style.pcss
```diff
@@ -0,0 +1,14 @@
+/**
+ * Styles specific to this block
+ */
+
+.wp-block-query-no-results {
+
+	/**
+	* Hack to get around query block displaying no results blocks
+	* on final page of results
```
dpellenwood, line 9: IIRC, this is a known bug in Gutenberg, right? It would be nice to reference the Issue or PR where they are addressing this so we can check in future versions if this is still needed.

dpellenwood (approved):
One suggested comment update, but this looks great, Geoff. Nice use of container queries!

wp-content/themes/core/blocks/core/lists/style.pcss
GeoffDusome: Correct yeah, `wp-site-blocks` is the wrapping class for all blocks on the FE & editor.

wp-content/themes/core/blocks/core/querynoresults/style.pcss
GeoffDusome: Apparently this was merged recently but I'm not seeing it fixed in 6.2.1 - maybe next release? https://github.com/WordPress/gutenberg/pull/53772

PR #103, opened 2023-09-12
[FEATURE] Gloablly Enqueued Core Block Styles

GeoffDusome:
## What does this do/fix?

- adds `$global_enqueue` variable to the `Block_Base` class to allow core block styles to be enqueued globally. 
- this solves some issues we've seen on a few projects where we've needed some core block custom styling to be available globally.
- currently I believe this is an issue where were duplicating core block markup for use in places where blocks aren't used. If we're not using the actual block, the assets don't get enqueued
- without this option, our only other option is duplicating styling rules
- @JMRhodes initially implemented this on QCHI. I also believe there _was_ a need on Comic-con for this (I'm not sure if there still is or not, but I believe they are duplicating styles in a few places to remedy). cc @Vinsanity @LayaTaal

### DIFF
--- a/wp-content/plugins/core/src/Blocks/Block_Base.php
+++ b/wp-content/plugins/core/src/Blocks/Block_Base.php
@@ -90,7 +90,7 @@ abstract class Block_Base {
 			wp_enqueue_style(
 				$this->get_block_handle(),
 				$src,
-				$this->get_block_dependencies(),
+				$args['dependencies'] ?? [],
 				$args['version'] ?? false,
 				'all'
 			);
--- a/wp-content/plugins/core/src/Blocks/Block_Base.php
+++ b/wp-content/plugins/core/src/Blocks/Block_Base.php
@@ -79,14 +79,17 @@ abstract class Block_Base {
 		$path     = $this->get_block_path();
 		$src_path = get_theme_file_path( "dist/blocks/$path/style-index.css" );
 		$src      = get_theme_file_uri( "dist/blocks/$path/style-index.css" );
+		$args     = $this->get_asset_file_args( get_theme_file_path( "dist/blocks/$path/index.asset.php" ) );
 
 		if ( ! file_exists( $src_path ) ) {
 			return;
 		}
 
-		if ( $this->global_enqueue ) {
-			$args = $this->get_asset_file_args( get_theme_file_path( "dist/blocks/$path/index.asset.php" ) );
+		// Add the style in the editor
+		add_editor_style( $src );
 
+		if ( $this->global_enqueue && ! is_admin() ) {
+			// Enqueue the style unconditionally on the public site
 			wp_enqueue_style(
 				$this->get_block_handle(),
 				$src,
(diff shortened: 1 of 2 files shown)


ChrisMKindred (approved):
Looks good to me.

wp-content/plugins/core/src/Blocks/Block_Base.php
```diff
@@ -83,7 +84,20 @@ public function enqueue_core_block_public_styles(): void {
...
+				$src,
+				$args['dependencies'] ?? [],
+				$args['version'] ?? false,
+				'all'
+			);
+		} else {
+			wp_add_inline_style( $handle, file_get_contents( $src_path ) );
+		}
+
 		add_editor_style( $src );
```
dpellenwood, line 101: A couple suggestions, if you'd be so kind:
```suggestion
		// Add the style in the editor
		add_editor_style( $src );

		if ( $this->global_enqueue ) {
			// Move `$args = ....` line to the top of the method with the other property definitions.
			// Even though it is only used conditionally, it's better to declare all setup vars in the same location.)
			// $args = $this->get_asset_file_args( get_theme_file_path( "dist/blocks/$path/index.asset.php" ) );
			// Enqueue the style unconditionally on the public site
			wp_enqueue_style(
				$this->get_block_handle(),
				$src,
				$args['dependencies'] ?? [],
				$args['version'] ?? false,
				'all'
			);
				
			return;
		}

		// Conditionally inline the style for the public site
		wp_add_inline_style( $handle, file_get_contents( $src_path ) );

```

dpellenwood (changes requested):
A couple questions:
* Do WP core styles for blocks used out side of the editor get added properly? If not, will that lead to issues where we've only got partial styles in that context?
* Have you tried leaving both the the `wp_enqueue_style()` and `wp_add_inline_style()` in place? Does the style get inserted twice in that situation?

Also, some changes to the PHP for clarity and future self, if you'd be so kind.

wp-content/plugins/core/src/Blocks/Block_Base.php
dpellenwood: Sorry for the confusion. My first question is not related to styled insert into the editor. It's specific to core block styles being rendered on the frontend if the block in question was not used in the block editor (post_content). For example, If I were to add a columns block (or use the columns block classes) in a site footer template--but not in the current posts's content--would WordPress's core styles for that columns block be inserted on the frontend or not?

I ask this because the `wp_add_inline_style()` is designed to attach additional (our custom) styles to the existing  styles for a block. So, as long as the core block styles are being rendered on the FE, then our custom styles should also be rendered.

On the flip side, if the core styles are _not_ being rendered on the front end for blocks that are not used in the post_content, then isn't there a chance that the styling of that block (columns in this case) will be incorrect?

GeoffDusome: I see what you're saying. I think the main issue we're trying to solve here is for the button block, which, for the most part, doesn't get affected by the blocks' core styles being missing because of how heavily we've overridden the block with our own styles in `theme.json` & custom block styles. I can see where issues would creep in if you're trying to use this on other blocks.

I'm willing to hold off on this PR for the next project to see if we run into this issue again and maybe we can come up with another solve. 

The specific example from comic-con was that they wanted to be able to style form submits & a few other unrelated button type things the same as the core buttons. They just ended up duplicating those styles but maybe that's how it _should_ be done and we leave core blocks alone to exist as their own thing? Maybe just be a bit better at documenting in-code if there are similar code blocks elsewhere?

GeoffDusome:
Closing - let's look at how we can better implement code like this in the next project.

PR #104, opened 2023-09-13
remove animation styles from base theme

GeoffDusome:
## What does this do/fix?

- removes animation script styles from base theme, they should have been removed previously but I missed them.


JMRhodes (approved):
👍

PR #106, opened 2023-09-25
Terms block php update

ChrisMKindred:
## What does this do/fix?

PHP Refactor for terms block.


GeoffDusome (approved):
🟢 👍 Thanks for tackling, Chris! 🌮

PR #108, opened 2023-10-10
PHPCS Cleanup

ChrisMKindred:
## What does this do/fix?

Upgrades default PHP Version for PHPCS to 8 to match our current minimum suggested PHP Version.
Adds escaping for `get_block_wrapper_attributes()` using `wp_kses_data()`.
Removes unused plugins from Composer.


dpellenwood (approved):
Looks good! Thanks for grabbing these fixes.

PR #109, opened 2023-10-16
WordPress 6.3.2 and Plugin Updates

ChrisMKindred:
## What does this do/fix?

Updates WordPress Core from 6.3.1 to 6.3.2
Updates the following plugins:
- advanced-custom-fields-pro to 6.2.1.1
- disable-emojis to 1.7.6
- limit-login-attempts-reloaded to 2.25.25
- seo-by-rank-math to 1.0.203

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,6 +75,21 @@
 		{
 			"type": "composer",
 			"url": "https://connect.advancedcustomfields.com"
+		},
+		{
+			"type": "package",
+			"package": {
+				"name": "block-editor-custom-alignments/block-editor-custom-alignments",
+				"version": "1.0.6",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -111,7 +96,6 @@
 	},
 	"require": {
 		"php": "^8.0|^8.1",
-		"block-editor-custom-alignments/block-editor-custom-alignments": "*",
```
dpellenwood, line 114: Curious why you're removing this (and the source block above)? I'm pretty sure we'll plan to use this on every project.

dpellenwood (approved):
Just one question re: the custom alignments.

PR #110, opened 2023-10-16
Adds Lighthouse CI GitHub Action

ChrisMKindred:
Due to a limitation in GitHub actions, this will need to be merged to the main branch before running the workflow as a full test.

## What does this do/fix?

This PR creates a Lighthouse CI Action that can be run via a workflow_dispatch. When the workflow runs, it will query one to many (line separated) of the URLs listed in the `urls` param.  The results will trigger an error or warning in the action UI based on the config settings in the `lighthouserc.json` file.  These settings should be customized for each production site based on the current performance. The full report is also added to the action artifacts, which can be disabled by changing the `uploadArtifacts` param to false.

This action aims to eventually run after deployments to test the site and trigger notifications if there are any performance, accessibility, SEO, or best practice degradations.

### DIFF
--- a/.github/lighthouse/lighthouserc.json
+++ b/.github/lighthouse/lighthouserc.json
@@ -1,0 +1,35 @@
+{
+    "ci": {
+        "assert": {
+            "assertions": {
+                "categories:performance": [
+                    "warn", {
+                        "aggregationMethod": "optimistic",
+                        "minScore": 0.8
+                    }
+                ],
+                "categories:accessibility": [
+                    "error", {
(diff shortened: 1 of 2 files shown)


dpellenwood (approved):
Looks good. I wonder if we should add some instructions to the project readme for this? Also: should this have a changelog entry?

ChrisMKindred:
@dpellenwood Yes, I will add both a readme/instructions and changelog but I need to get it merged into main first to make sure that everything runs.

PR #111, opened 2023-10-16
Adds Docs for Actions

ChrisMKindred:
## What does this do/fix?

Creates a docs folder and adds high-level docs for the GitHub Actions.


docs/actions.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Static Analysis and Testing
+
+There are a series of workflows that must pass in order to allow a pull request to be merged.  These are highlighted in the `workflows.yml` file and include the coding standards, static analysis, linting, and the php test suite using [Slic](https://github.com/stellarwp/slic).
+
+## Lighthouse Testing
+
+We are using the [Lighthouse CI](https://github.com/treosh/lighthouse-ci-action/tree/main) for testing list of production urls that we would like to run lighthouse tests on and stores the results as artifacts of the action. There are parameters that we set in the [lighthouserc.json](../.github/lighthouse/lighthouserc.json) file allowing us to set the minimum values for each of the lighthouse matrix. There are minimum values set as a baseline but each value should be updated once a project is live in order to track that updates made do not effect the results over time.
```
dpellenwood, line 21: I'd probably make it a little clearer that the config does need to be set specifically for each project. (That this action won't work in a project as-is.) Otherwise, this is excellent. Thanks for writing it up!

dpellenwood (approved):
One suggestion, but looks great!

PR #112, opened 2023-10-17
[FIX] Terms Block

GeoffDusome:
## What does this do/fix?

I made these same edits on Soros (in a draft PR), so whatever we do here I'll pull over there & into Moose Resources

- uses block context to get post type for the current post (in both ind post & query block contexts) in order to pull in only post types that relate to the current post for better editor experience
- displays placeholder text if no terms are returned & you're viewing the block in the editor; if you're not in the editor and no terms return, an HTML comment is displayed instead.
- I don't know that checking the request is the best way to check if we're in the editor, but honestly, I couldn't find another way and I was getting frustrated with it. Feel free to take a look. Maybe I missed something.

## QA

Links to relevant issues
- [Link to Issue](https://tribe.slack.com/archives/C04N3SZ7QVD/p1697121910892039)


wp-content/themes/core/blocks/tribe/terms/render.php
```diff
@@ -16,7 +16,16 @@
...
+	if ( strpos( $_SERVER['REQUEST_URI'], 'wp-admin' ) !== false || strpos( $_SERVER['REQUEST_URI'], 'wp-json' ) !== false ) {
+		echo '<ul class="wp-block-tribe-terms__list">';
+		echo '<li class="wp-block-tribe-terms__term">';
+		echo '<span class="wp-block-tribe-terms__link t-category">'. esc_html__( 'Terms Display Here', 'tribe' ) .'</span>';
+		echo '</li>';
+		echo '</ul>';
+	} else {
+		echo '<!-- Terms block: No terms to list. -->';
+	}
+
```
dpellenwood, line 28: Curious: Do we _need_ to render something different in the editor than the admin? I wonder if we render just don't render anything, always. If it needs to be visually distinguished in the editor, maybe we add an empty results css class to the container and use editor-specific css?

wp-content/themes/core/blocks/tribe/terms/block.json
```diff
@@ -28,6 +28,7 @@
 	"styles": [
 		{ "name": "pills", "label": "Pills" }
 	],
+	"usesContext": [ "postType" ],
```
dpellenwood, line 31: 😍

wp-content/themes/core/blocks/tribe/terms/edit.js
```diff
@@ -26,13 +26,15 @@ import ServerSideRender from '@wordpress/server-side-render';
...
+	context,
+	isSelected,
+} ) {
 	const blockProps = useBlockProps();
 	const { taxonomyToUse, onlyPrimaryTerm, hasLinks } = attributes;
-
-	const postType = useSelect( ( select ) =>
-		select( 'core/editor' ).getCurrentPostType()
-	);
+	const { postType } = context;
```
dpellenwood, line 37: 🤌

dpellenwood (commented):
Love the improvements. I left my thoughts on the admin detection. Let me know what you think.

wp-content/themes/core/blocks/tribe/terms/render.php
GeoffDusome: Good call - I ended up taking this route in my newest commit. I'm checking if the block is empty in the editor and adding a pseudo element styled how it would look in the FE. I like this approach more, it's just not possible to translate the string (which I don't know is a problem, either). Let me know how you like this approach vs Chris' separate PR.

dpellenwood (approved):
Add a changelog entry and then you're good to merge this when you have a minute, @GeoffDusome . Thanks for your patience with us.

PR #113, opened 2023-10-17
Rest Request and Fake the Term

ChrisMKindred:
## What does this do/fix?

If we check the `REST_REQUEST`, it will tell us if we are using the editor.  Then, we can just fake the terms to show the user an item instead of duplicating the HTML. 

Give this a test and let me know what you think.

### DIFF
--- a/wp-content/plugins/core/src/Blocks/Terms_Block.php
+++ b/wp-content/plugins/core/src/Blocks/Terms_Block.php
@@ -28,15 +28,36 @@ class Terms_Block {
 	 */
 	public function get_the_terms(): array {
 		if ( $this->only_primary_term ) {
-			$term = $this->get_primary_term( get_the_ID(), $this->taxonomy );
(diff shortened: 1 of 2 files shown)


GeoffDusome:
@ChrisMKindred I like this approach as well because it allows us to translate the string. In my most recent commit in the main branch I took a CSS approach. I'm going to let Dave weigh in with which approach he likes better and maybe we can merge them. Thanks for taking the time!

ChrisMKindred:
@GeoffDusome, I think @dpellenwood is doing some digging as well. Between the 3 of us, we will have a couple of options.  I am not sure what I think about the text with CSS. The translation is one good reason, but also in terms of management, as a dev you don't always think to look at the CSS for actual text and it could cause some additional time to troubleshoot in the future.

ChrisMKindred:
Closed in favor of #114

PR #114, opened 2023-10-18
Fix/terms block taxonomies no terms

dpellenwood:
## What does this do/fix?

* Adds a no terms message in the block editor (based on Chris's REST_REQUEST approach). 
* Simplifies the render to use the same markup for all contexts where there are terms (both single & multiple)


wp-content/themes/core/blocks/tribe/terms/render.php
```diff
@@ -15,45 +15,35 @@
 
 echo '<div ' .  wp_kses_data( get_block_wrapper_attributes() ) . '>';
 
-if ( 0 === count( $terms_block_terms ) ) {
-	echo '<!-- Terms block: No terms to list. -->';
-	echo '</div>';
-
-	return;
+// No terms and we're in the block editor
+if ( 0 === count( $terms_block_terms ) && ( defined( 'REST_REQUEST' ) && REST_REQUEST ) ) {
```
dpellenwood, line 19: I think we should probably extract `( defined( 'REST_REQUEST' ) && REST_REQUEST ) )` into a utility/helper somewhere more global. It feels weird to check like this in the context. But, food for thought.

wp-content/themes/core/blocks/tribe/terms/render.php
```diff
@@ -15,45 +15,35 @@
...
 }
 
-if ( count( $terms_block_terms ) > 1 ) {
-	echo  '<ul class="wp-block-tribe-terms__list">';
-}
+if ( count( $terms_block_terms ) > 0 ) {
+	echo '<ul class="wp-block-tribe-terms__list">';
 
-foreach ( $terms_block_terms as $term ) {
-	if ( count( $terms_block_terms ) > 1 ) {
```
dpellenwood, line 30: I don't think it's worth switching the markup depending on if there is only a single term displayed. The block itself is a "terms block" which infers there could be one or more.

dpellenwood (commented):
Left a couple notes on my own code for conversation.

wp-content/themes/core/blocks/tribe/terms/render.php
```diff
@@ -15,45 +15,35 @@
 
 echo '<div ' .  wp_kses_data( get_block_wrapper_attributes() ) . '>';
 
-if ( 0 === count( $terms_block_terms ) ) {
-	echo '<!-- Terms block: No terms to list. -->';
```
dpellenwood, line 19: We render the block wrapper div regardless, so I don't see the comment as necessary.

wp-content/themes/core/blocks/tribe/terms/style.pcss
```diff
@@ -52,3 +52,14 @@
...
+/* -----------------------------------------------------------------------
+ * Terms Block - Empty Terms List
+ * ----------------------------------------------------------------------- */
+
+.wp-block-tribe-terms__empty-msg {
+
+	.editor-styles-wrapper & {
+		color: var(--color-neutral-60);
+	}
+}
```
GeoffDusome, line 65: ❓ Should we add the `t-tag` mixin here so the styling shows correctly depending on the style chosen or is it ok that this displays differently?

GeoffDusome (approved):
🟢 This works for me - kinda bridges both solutions. Just one question.

wp-content/themes/core/blocks/tribe/terms/style.pcss
dpellenwood: Good catch. Updated.

dpellenwood:
Made a few additional changes to reduce the style bleed, and align with how core renders the Post Tags block in the editor when no terms are assigned.

wp-content/themes/core/blocks/tribe/terms/editor.pcss
```diff
@@ -1,16 +1,14 @@
-.wp-block-tribe-terms {
+.wp-block-tribe-terms {}
```
GeoffDusome, line 1: ❓ Do we need this empty selector?

GeoffDusome (approved):
🟢 Just one question, other than that good to go!

ChrisMKindred (approved):
👍

wp-content/themes/core/blocks/tribe/terms/editor.pcss
dpellenwood: Good catch. Removed.

PR #119, opened 2023-11-30
[FIX] Mobile Column Stacking Order

GeoffDusome:
## What does this do/fix?

- fixes an issue where an extra div was being added to the Column block causing any Columns block layout other than 50/50 to not display correctly.


dpellenwood (approved):
🐠 💨

PR #120, opened 2023-12-04
Adds override for triggering xdebug with using WP-CLI

ChrisMKindred:
## What does this do/fix?

Adds the ability to trigger xdebug with wp-cli when using lando. Requires `lando rebuild` to go into effect on current projects.


dpellenwood (approved):
🚀 💨

PR #126, opened 2024-03-05
[MOOSE-86] Allow `view.js` files for blocks

GeoffDusome:
## What does this do/fix?

- rewrites `blockEntryPoints` function to allow `view.js` files to generate within blocks
- since all entry point loops are the same, we can just use one array & loop to handle all cases

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-86)

Screenshots/video:
- `files` array upon running `npm run dist`: http://p.tri.be/i/7mV7gs (`console.log` has been removed)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -62,34 +62,31 @@ const assetEntryPoints = () => {
  */
 const blockEntryPoints = () => {
 	/**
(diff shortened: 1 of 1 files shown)


dpellenwood (approved):
🚚 💨

PR #128, opened 2024-03-06
[MOOSE-91/109] Migrate "Create WP Controls" script

GeoffDusome:
## Why was this change made?

Prior to this change, every core block that needed additional controls added (say, if we needed a control to make it easy to add/remove a class on a block) would need a 150+ line long JS file just to add one simple control. The purpose of this change is to make adding these controls more efficient for developers, create less points of failure, reduce the overall code footprint, and to allow easy extensibility of the functionality (add different control types) included in the script.

## What does this do/fix?

- Migrates a script created for Panopto that handles creating WP controls for core blocks (can also be used for custom blocks that need sidebar controls). 
- Script has the ability to add `ToggleControl`, `NumberControl`, and `SelectControl` components to cover all Moose usages of the script. Other controls are not currently supported, but can be easily added based on need.
- Adds the following _new_ controls
    - Image block: "Remove Margins" control. Toggle to remove vertical margins from the image block.
    - Image block: "Width Fills Container" control. Toggle to ensure the image fills the width of its' container.
- Based on DE feedback, hyphenated text has been removed globally from the reset.
- Aspect ratio helpers have been moved to utilities file within the media folder. I've added an extra note to these utilities to ensure they are only used when you need to add an aspect ratio that doesn't exist within WordPress (or you need to aspect ratio something that doesn't have aspect ratio controls by default).
- To go along with the new Image block controls, I've added new utility classes for overriding default margins on blocks (these are automatically applied via the new Image block control). 
- Fixed issue with Columns block where it wasn't using the proper spacing variable for the column gap.

## QA

Links to relevant issues
- [MOOSE-91](https://moderntribe.atlassian.net/browse/MOOSE-91)
- [MOOSE-109](https://moderntribe.atlassian.net/browse/MOOSE-109)

Screenshots/video:
- [Link to Loom](https://www.loom.com/share/c28e3c47128a4380b95d2a5a22c3c912)

[MOOSE-91]: https://moderntribe.atlassian.net/browse/MOOSE-91?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


LayaTaal:
@GeoffDusome I appreciate you working on an API to make creating editor controls a little easier for us. I am concerned that this would add extra complexity that isn't necessary though. Maintaining this as we include more and more use cases could be challenging and there are a lot of areas to protect from errors. I worry we would spend more time fixing bugs than we would just using the standard WordPress API to create controls.

dpellenwood:
Not ignoring this PR @GeoffDusome . I just need more time to think about it a bit.

GeoffDusome:
@dpellenwood Hey, I realized I didn't provide any details on _why_ I made this change. I've added a section in the PR description above to help align expectations in what my intentions were with this change. Hopefully this help gives a bit of background on why the change was made.

GeoffDusome:
@jamiepaul @bjcooper @LayaTaal @Vinsanity Hey all, can I get your opinions on this PR? Would it be helpful to you to have a quick way to add controls to core blocks?

Vinsanity:
@GeoffDusome I love the effort here and concept. I can see both sides of this argument so here's my feedback.
- While this would be specific to core blocks for now, I think some of the options being added by default are project/design specific. Balanced text, and Image width and margins for example. This is something I'd rather define in the CSS and CSS vars. Image margins for instance, while you can toggle that, what's stopping them from adding the margin from dimensions? if we're removing that with theme.json, are we restricting their ability to create the page they want?

GeoffDusome:
@Vinsanity @LayaTaal Hey - made some updates here to remove some of the more opinionated controls that I added. I agree that we don't need these in every project. They were only added because we did it on one project, and design thought we would continue using that through other projects, which didn't end up happening. 

This PR is much more tame now, only adding the Image block controls (which are still super helpful to avoid needing custom code to remove image margins in a pattern & setting images to 100% width). 

Let me know if this feels like a better direction, just allowing developers to create controls for core blocks much quicker.

wp-content/themes/core/assets/pcss/media/_utilities.pcss
```diff
@@ -0,0 +1,34 @@
...
+	aspect-ratio: 1 / 1;
+}
+
+.s-aspect-ratio-4-3 {
+	aspect-ratio: 4 / 3;
+}
+
+.s-aspect-ratio-16-9 {
+	aspect-ratio: 16 / 9;
+}
```
dpellenwood, line 34: Since these are no longer spacing selectors, probably should remove the `s-` prefix, yes?

dpellenwood:
Thanks for your thoughtful approach here, @GeoffDusome . I think the utility script for scaffolding controls is a nice addition.  I'd suggest it should be documented really clearly so that other know how to use it quickly without having to reference your implementations. Maybe evcen an overvirew and example control build with it in a readme file in the utils folder? 

That being said, I'm not sure I'm on board with the (subset of) default controls that are created here. My thoughts:
1. Text Balance - I discussed this with Sarah back a while ago and we came to the conclusion that it would be best if design  specifies the contexts where that css feature is to be used. This reduces the chance that it gets added in some contexts and not others. Effectively, design would like more stringent control over its usage throughout a project and by not creating a UI control for it there is less training/editor knowledge required.
1. Image margins. This, too feels like a situation where we should programmatically control the layout, imho. From the surface it seems we just need to add a css rule that removes the image margins when the image appears within certain contexts. Is that a possibility? If we give editors a control to toggle this on & off, there's bound to be situations where both margined and un-margined images end up appearing in the content, which leads to inconsistencies.

Again, thanks for tackling this. I really appreciate your input on incremental improvements like this!

wp-content/themes/core/assets/pcss/global/reset.pcss
```diff
@@ -34,15 +34,6 @@ figcaption {
...
 
-/**
- * Avoid text overflows.
- */
-p,
-:--heading {
-	overflow-wrap: break-word;
-	hyphens: auto;
-}
-
```
dpellenwood, line 45: 👍

dpellenwood (commented):
Thanks for your thoughtful approach here, @GeoffDusome . I think the utility script for scaffolding controls is a nice addition.  I'd suggest it should be documented really clearly so that other know how to use it quickly without having to reference your implementations. Maybe evcen an overvirew and example control build with it in a readme file in the utils folder? 

That being said, I'm not sure I'm on board with the (subset of) default controls that are created here. My thoughts:
1. Text Balance - I discussed this with Sarah back a while ago and we came to the conclusion that it would be best if design  specifies the contexts where that css feature is to be used. This reduces the chance that it gets added in some contexts and not others. Effectively, design would like more stringent control over its usage throughout a project and by not creating a UI control for it there is less training/editor knowledge required.
1. Image margins - This, too feels like a situation where we should programmatically control the layout, imho. From the surface it seems we just need to add a css rule that removes the image margins when the image appears within certain contexts. Is that a possibility? If we give editors a control to toggle this on & off, there's bound to be situations where both margined and un-margined images end up appearing in the content, which leads to inconsistencies.
2. 1. Image Full Width - I worry that this will lead to editors incorrectly using images that are too small and then just scaling them to fit rather than being required to insert images that properly fill (at resolution) a given container.

Again, thanks for tackling this. I really appreciate your input on incremental improvements like this!

wp-content/themes/core/assets/pcss/media/_utilities.pcss
GeoffDusome: Removed prefix and changed aspect ratios to ratios that do not exist currently in the aspect ratio picker on the image block.

GeoffDusome:
@dpellenwood I removed the Image block controls. The balanced text controls on the Heading & Paragraph blocks were removed previously. I agree that this type of thing can be implemented on a per site basis.

dpellenwood (approved):
🟢 Looks great! Thanks for seeing this one through!

PR #131, opened 2024-03-12
[FIX] Terms Block

GeoffDusome:
## What does this do/fix?

- fixes issue where the Terms block would error when no post ID was provided to it
- sets defaults for attributes in the Terms block - should prevent any "missing attribute" issues
- since the `taxonomies` array is gathered by an async request, we should be checking if it exists before we loop over nothing.

## QA

Links to relevant issues
- [Link to Panopto Slack thread](https://tribe.slack.com/archives/C061UC7B2F9/p1710250320818599)


dpellenwood (approved):
🎾 💨

PR #134, opened 2024-03-15
[MOOSE-77] Block Templates

dpellenwood:
## What does this do/fix?

- adds two new `create-block` variants (static, dynamic)
- a "static" block is a JS rendered block
- a "dynamic" block is a PHP rendered block using `ServerSideRender`
- both types of blocks add an example attribute and matching `TextControl` to update the attribute - keep in mind that this is just an example of how to handle attributes inside of a block. There are many different ways an attribute can be updated within a block. The general rule I take is if it's a block setting in a static block, or the block is a dynamic block, use sidebar (or toolbar, depending on need) controls - else, we can do some cool stuff where we use inline controls to update attributes.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-77)

Screenshots/video:
- [Link to Video](https://www.loom.com/share/dc697e37432c4781a1772531ce379b60)


GeoffDusome:
@dpellenwood I was able to get variants working by adding them to the exports in the templates' `index.js` file. I was also able to add some default values. I also added the template call to the main `create-block` script call, so it doesn't need to be typed. 

Variants can be chosen through the `--variant` param: `npm run create-block -- --variant dynamic` or, the script will prompt you to choose, if you don't type it (this is just easier lol). 

My assumption here is that we could get as complex or simple as we want to go. As we start building more blocks, it may become apparently that we do "x" every time, and it would just make sense to include that in the template. Adding more variants is always an option as well (I'm thinking stuff like sidebar controls vs none, etc).

I'll let you tell me how far you'd like to go to start with. I'd personally love to clear out some of those large comments.

dev/templates/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "tribe-block-template",
+  "version": "1.0.0",
+  "description": "Custom Block Template for the wp create block script.",
```
dpellenwood, line 4:
```suggestion
  "description": "Custom block templates for the WP create-block script to use.",
```

dpellenwood (commented):
I left a few suggestions. I agree that it would be nice to clean up the boilerplate comments quite a bit, set some defaults that we'd actually work from, and then introduce a few variations. It would be cool to be able to generate a variation for a core block that doesn't have a default `block.json` file, but I don't know if that's possible.

This is an area where we can kind of trial & error a bit until we get exactly what we want dialed in. Feel free to have fun with it. :)

package.json
```diff
@@ -54,8 +54,8 @@
 		"lint:js:fix": "wp-scripts lint-js \"$npm_package_config_coreThemeDir/**/*.js\" --fix",
 		"lint:configs": "wp-scripts lint-js \"./*.js\"",
 		"lint:configs:fix": "wp-scripts lint-js \"./*.js\" --fix",
-		"lint:pkg-json": "wp-scripts lint-pkg-json",
-		"create-block": "cd \"$npm_package_config_coreThemeBlocksDir/tribe\" && npx @wordpress/create-block --no-plugin --namespace tribe",
+		"lint:pkg-json": "wp-scripts lint-pkg-json \"./package.json\"",
+		"create-block": "cd \"$npm_package_config_coreThemeBlocksDir/tribe\" && npx @wordpress/create-block --no-plugin --namespace tribe --template ../../../../../dev/templates/",
```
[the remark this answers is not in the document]
GeoffDusome: @dpellenwood This was the only thing I couldn't update, based on our chat in Slack. It's still a bit better with the relative URL in the config property, but obviously still kinda gross. Let me know if you think we could do it any other way (I couldn't figure our any other way so I went back to a way I knew was working).

dev/templates/block/editor.pcss.mustache
```diff
@@ -0,0 +1,9 @@
+/* -------------------------------------------------------------------------
+ *
+ * Block - {{title}} - Styles - Editor Only
+ *
+ * ------------------------------------------------------------------------- */
+
+{{#isDynamicVariant}}section{{/isDynamicVariant}}.wp-block-{{namespace}}-{{slug}} {
```
GeoffDusome, line 7: Just a note here that this prefixed `section` is due to dynamic blocks rendering the wrapper div + the wrapping `section` in the render template (`render.php`). We're only wanting to target the section (the actual block) in this case.

GeoffDusome:
@jamiepaul @LayaTaal @Vinsanity Any thoughts on this functionality? Do you think we can merge just handling static & dynamic blocks?

I know we talked about it a bit ago and I said I was going to try to also do core blocks, but thinking about it now, every time you run this code, it's going to generate a `block.json` file, which we don't need for our core block directories as it is now. Currently it's easier to just duplicate a folder there, rather than have to worry about deleting files in the generated code. My other thought here is that because of the way our namespacing needs to work, it would be pretty hard to generate the required types of "slug" we need (the Post Author Name block has 3 different strings: "Post_Author_Name", "postauthorname", and "post-author-name"). I recommend sticking with manual entry for core blocks. We're getting to the point now where it's very rare we need to add a new one anyway.

LayaTaal:
I think this is a great addition! I love that you can just run `npm run create_block` and let the prompts guide you. A few questions I have:

- Can we include the textdomain in the template? It would be nice the same text domain for every block.
- Could we include a README along with this. A link to WordPress docs on create-block templates could be helpful to include.
- Do we still need the `example` block? It feels like it is easy enough to just use this that we don't really need the example block any longer.

GeoffDusome:
@LayaTaal Did a bit of research into this and turns out because the script uses the same code for external templates as it does for creating a custom block plugin, it uses the block slug as the `textdomain` attribute in `block.json` at all times. This is one of the gotchas here and we'll have to update that one field manually. Small price to pay I guess.

I added documentation under `./docs` for block templates (file named `block-templates.md`. Should be a good jumping off point for understanding the script.

Good call, I've removed the example block.

GeoffDusome:
@jamiepaul @Vinsanity @bjcooper @Vinsanity Hey all - this one is good for review as well - please read the feedback previously given for some insight into the current state of the PR. Hoping to get this merged so everyone can use it on their projects.

Vinsanity (approved):
🟢 This is lovely! Tested it a few times and tried to give it some errors which were all recognized and prompted. It's unfortunate that the textdomain is locked into the WP scripting but maybe in the future they'll loosen that up.

LayaTaal (approved):

PR #135, opened 2024-03-25
Fix/remove mtribe composer util

dpellenwood:
## What does this do/fix?
- Remove Gravity Forms as a composer dependency and the respective mtribe.site composer utility. Gravity Forms should be added directly to a project repo when required.

[Related Slack conversation.](https://tribe.slack.com/archives/C059WEJF9PB/p1696542199114139?thread_ts=1696525529.011279&cid=C059WEJF9PB)

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,7 +1,0 @@
-# Find API keys in 1Password, in the Square One vault under "Square One .env"
-# Direct 1password link: https://m.tri.be/soenv10
-#
-# ACF Key has been move to the auth.json file with the license being used for
-# the username and the full site url for the password.
-WP_PLUGIN_GF_KEY=''
-WP_PLUGIN_GF_TOKEN=''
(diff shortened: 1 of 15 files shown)


JMRhodes (approved):
👍

jamiepaul (changes requested):
🟠 README.md still includes instructions related to the `.env` and Gravity Forms initial setup. We should probably remove those notes as well.

MlKilderkin:
> 🟠 README.md still includes instructions related to the `.env` and Gravity Forms initial setup. We should probably remove those notes as well.

I believe we can also cleanup .env.sample or even delete it at all since we don't need it now

MlKilderkin (approved):

dpellenwood:
> 🟠 README.md still includes instructions related to the `.env` and Gravity Forms initial setup. We should probably remove those notes as well.

Good catch, @jamiepaul & @MlKilderkin . I went ahead and removed the references to the .env file from the docs as well as the sample and related deployment actions. Give this another look?

MlKilderkin (approved):

jamiepaul (approved):
🟢 👍🏻

PR #136, opened 2024-04-03
[FIX] Visual Editor Post Title

GeoffDusome:
## What does this do/fix?

- the post title update we made had a few issues when adding a new post
- this change moves to using absolute positioning for the "Page Title" pseudo element to avoid changing the layout structure of the post title field by using flex as we were before.
- this change also adds a variable in `layout/_variables.pcss` for the "Grid" width set in `theme.json`
- updates version of Block Editor Custom Alignments plugin to `1.0.7`
- updates layout variables to use default grid width variable defined by the new version of the plugin

## QA

Links to relevant issues
- N/A - noticed in DQA on F&D

Screenshots/video:
- Adding a new post should now look like this: http://p.tri.be/i/vm5XjE


wp-content/themes/core/assets/pcss/layout/_variables.pcss
```diff
@@ -12,4 +12,7 @@ body {
 	--grid-margin: var(--wp--custom--spacing--grid-margin);
 	--grid-gutter: var(--wp--custom--spacing--grid-gutter);
 	--grid-max-width: var(--wp--style--global--wide-size);
+
+	/* this should match the "Grid" size defined in theme.json */
+	--grid-max-grid-width: 1680px;
```
jamiepaul, line 17: 🔴  could we differentiate the variable names better between `--grid-max-width` and `--grid-max-grid-width`? Maybe it makes the most sense to rename `--grid-max-width` to `--grid-wide-width` or something?

jamiepaul (changes requested):
Post/Page title fix looks good, but just a small change needed on the variable names

wp-content/themes/core/assets/pcss/layout/_variables.pcss
```diff
@@ -11,5 +11,8 @@
 body {
 	--grid-margin: var(--wp--custom--spacing--grid-margin);
 	--grid-gutter: var(--wp--custom--spacing--grid-gutter);
-	--grid-max-width: var(--wp--style--global--wide-size);
+	--grid-wide-width: var(--wp--style--global--wide-size);
+
+	/* this should match the "Grid" size defined in theme.json */
+	--grid-max-width: 1680px;
```
dpellenwood, line 17: I wonder if we should programmatically create some css custom props based on the sizes we define in the tribe plugin?

At least for now, maybe name this something like `--tribe--style--global--grid-size` so it matches Core's global style names, but uses a custom namespace? I suppose to match the plugin name something like `--beca--style...` would make sense.

I'd also make this comment a little clearer that these are the sizes defined in the `_experimentalLayout` section as used by our plugin.

dpellenwood (changes requested):
Left my thoughts on the variable naming here. `--grid-max-width` has a ton of connotations for us old SquareOne devs make make it almost useless. :melting:

GeoffDusome:
@jamiepaul @Vinsanity @LayaTaal @bjcooper Hey all, I've updated the Block Editor Custom Alignments plugin to use CSS variables for each defined width in your `theme.json`. I've updated this PR to reflect that change for the editor post titles. If I could get a few reviews / thoughts we can get this merged in!

jamiepaul (approved):
🟢 👍🏻

wp-content/themes/core/assets/pcss/global/editor.pcss
```diff
@@ -32,6 +31,9 @@
 
 			@mixin t-body;
 			content: "Page Title";
```
Vinsanity, line 33: @GeoffDusome something to consider for a future update maybe. For localization reasons, can we hook into the page title and inject an attribute with a localized string? end result being `data-title="Localized Page Title"` and then use `content: attr(data-title);`

Vinsanity (commented):
LGTM. 1 comment to review. 🟢

wp-content/themes/core/assets/pcss/global/editor.pcss
GeoffDusome: Short answer: It looks like it's not currently possible to do this through the use of any hooks. I could of course always use JS to find the element and manually set the attribute, but this feels gross. My suggestion for sites that require localization - remove the pseudo element and update the padding to match top & bottom. I think it's obvious enough what the element is without the helper label.

PR #139, opened 2024-05-03
Patterns: Consistency; Post Pattern Picker

GeoffDusome:
## What is the background for this change?

On a few sites now, we've been switching to use the pattern picker for posts (http://p.tri.be/i/xB7gRq, this displays when you go to add a new post) rather than automatically injecting the main post pattern into the page. Having to switch to using the pattern picker is a hassle if we're going to do it for every project, so I'm proposing that we move to this as a standard. This should allow a more simple transition to different post templates being available to an editor.

Also, our usage of `Inserter:` in the pattern templates was fairly inconsistent. I've removed `Inserter:` from any pattern where it was set to `yes`, as this is the default. We should only be using `Inserter:` if we're setting the value to `no`.

## What does this do/fix?

- updates consistency of the usage of `Inserter:` in pattern definitions

Updated 05/09/24:

- updates `Post` pattern definition to get called from the pattern selector
- updates `Post` pattern to bring it up to modern usage (Terms block, Wide Featured Image, Grid width Related Posts)
- updates `Search Result Post Card` pattern to remove double heading within card
- updates `Search` template to reflect new card usage
- removes utility that's no longer used in core Moose (and frankly, shouldn't be used)
- adds `Post Split Header` pattern that should act as an example of an alternative post pattern. Both this new pattern and the original `Post` pattern should show in the pattern picker. The `Post` pattern remains the default: http://p.tri.be/i/J3NRaV
- fixes issue with the Columns block styles where we were using the wrong variable for the vertical gap.


jamiepaul (approved):
🟢 👍🏻  I like this change- it's nice to have a choice to start from the template or not

Vinsanity (changes requested):
The update here makes sense. However, we've recently had clients making poor decisions when it comes to content entry. I feel we need to be more opinionated here. For that reason alone I'm suggesting the following change.

1. Add a Config and Subscriber back into the core plugin under Post_Type/Post. We should be forcing the layout provided by design to prevent content editors from making mistakes.
2. If they need any additional pattern layouts, we can certainly use the Inserter default to give them those options.
3. Maybe make an example post-alternate pattern in case we have multiple layouts.

FWIW, even when you define a pattern in that config, if you choose the option for a different starting point, it replaces the default.

GeoffDusome:
@Vinsanity made some updates. Noted under the "Updated 05/09/24" heading in the PR description. 

I'm hesitant to remove the "pattern picker always loads by default when creating a new post" because otherwise new developers aren't going to know that functionality even exists. If we feel strongly though, I can remove the `Post Split Header` pattern and remove the pattern selector when creating a new post entirely and we'll just have to remember & remind others that that's always an option.

PR #142, opened 2024-06-07
[DO NOT MERGE] Custom Query Loop Block

GeoffDusome:
Pushing Query Loop Block up for testing.


dpellenwood:
@GeoffDusome , I'm going to close this PR and [the respective Jira ticket](https://moderntribe.atlassian.net/browse/MOOSE-135). This work was **_exactly_** what we needed to spur a conversation with design about how this should work. It's awesome work. But, design is backed up a bit and the approach we decided on in our working session will require a new ticket/scope for dev anyway. Rather than keep these languishing for ever, I wan to close for now and come back to it after the scope & design is finalized. Thanks for taking the initial stab at this! 🌮

dpellenwood:
@GeoffDusome , I'm going to close this PR and [the respective Jira ticket](https://moderntribe.atlassian.net/browse/MOOSE-135). This work was **_exactly_** what we needed to spur a conversation with design about how this should work. It's awesome work. But, design is backed up a bit and the approach we decided on in our working session will require a new ticket/scope for dev anyway. Rather than keep these languishing for ever, I wan to close for now and come back to it after the scope & design is finalized. Thanks for taking the initial stab at this! 🌮

PR #145, opened 2024-06-21
Feature/action workflows cleanup

dpellenwood:
## What does this do/fix?
Minor improvements to some of the GitHub Action workflows:
* Renamed and added a finish job to the Dokku Deploy App workflow so that it doesn't fail when all 3 app jobs are skipped.
* Renamed the code quality workflow from "Workflow" to "Code Quality Checks" and renamed the file accordingly.


JMRhodes:
@dpellenwood probably a separate ticket, but as long as this is workflow related it might be worth mentioning that we're using some outdated actions in a few places, which is resulting in some version warnings during execution:
- https://github.com/moderntribe/moose/actions/runs/9617664104
- http://p.tri.be/i/Gxm0pc

dpellenwood:
> @dpellenwood probably a separate ticket, but as long as this is workflow related it might be worth mentioning that we're using some outdated actions in a few places, which is resulting in some version warnings during execution:
> 
> * https://github.com/moderntribe/moose/actions/runs/9617664104
> * http://p.tri.be/i/Gxm0pc

Good catch. Looks like we need to update *all* the GH checkout, use-node & related actions to a new version. I'll give that a go here, too.

PR #148, opened 2024-06-24
[MOOSE-132] Block Bindings API

GeoffDusome:
## What is the background of this update?

The Block Bindings API is gaining traction and slowly getting more features towards having an UI to use to set bindings without needing to know the specific binding you're wanting to target. This currently is not the case, so we're stuck adding examples to the binding class in a comment. 

Adding an interface to create the bindings makes adding bindings relatively simple. Currently, we do not have a usage for the Block Binding API in Moose that involves more than simple one line text applications so I do question if it makes sense to use the Block Binding API here over a custom dynamic block or even an ACF block.

I also question the usage of the Block Binding API within Moose in its current state and in the future. I do not see a world where the Block Binding API is _more_ useful than an ACF block would be. Hopefully someone else has an idea here but it feels very too early in the development of the Block Binding API to be using it for much more than these one-liners. 

## What does this do/fix?

- currently branched off of a branch that's using WP 6.6 for the specific related updates to the block bindings API. These updates do not depend on those, but there have been updates in WP 6.6 that make the bindings API slightly better.
- moves single line text blocks to block bindings API. 
- updates search patterns to reflect new usage

## Quick Implementation for Testing

Post Permalink:

```html
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"tribe/post-permalink"}}}} -->
<p>Post Permalink Placeholder</p>
<!-- /wp:paragraph -->
```

Post Type Name:

```html
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"tribe/post-type-name"}}}} -->
<p>Post Type Name Placeholder</p>
<!-- /wp:paragraph -->
```

Query Results Count:

```html
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"tribe/query-results-count"}}}} -->
<p>Query Results Count Placeholder</p>
<!-- /wp:paragraph -->
```

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-132)

Screenshots/video:
- Coming soon...


GeoffDusome:
Closing in favor of #150

PR #150, opened 2024-06-26
[MOOSE-132] Block Binding API Support

GeoffDusome:
## What is the background of this update?

The Block Bindings API is gaining traction and slowly getting more features towards having an UI to use to set bindings without needing to know the specific binding you're wanting to target. This currently is not the case, so we're stuck adding examples to the binding class in a comment. 

Adding an interface to create the bindings makes adding bindings relatively simple. Currently, we do not have a usage for the Block Binding API in Moose that involves more than simple one line text applications so I do question if it makes sense to use the Block Binding API here over a custom dynamic block or even an ACF block.

I also question the usage of the Block Binding API within Moose in its current state and in the future. I do not see a world where the Block Binding API is _more_ useful than an ACF block would be. Hopefully someone else has an idea here but it feels very too early in the development of the Block Binding API to be using it for much more than these one-liners. 

## What does this do/fix?

- currently branched off of a branch that's using WP 6.6 for the specific related updates to the block bindings API. These updates do not depend on those, but there have been updates in WP 6.6 that make the bindings API slightly better.
- moves single line text blocks to block bindings API. 
- updates search patterns to reflect new usage

## Quick Implementation for Testing

Post Permalink:

```html
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"tribe/post-permalink"}}}} -->
<p>Post Permalink Placeholder</p>
<!-- /wp:paragraph -->
```

Post Type Name:

```html
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"tribe/post-type-name"}}}} -->
<p>Post Type Name Placeholder</p>
<!-- /wp:paragraph -->
```

Query Results Count:

```html
<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"tribe/query-results-count"}}}} -->
<p>Query Results Count Placeholder</p>
<!-- /wp:paragraph -->
```

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-132)

Screenshots/video:
- Coming soon...


wp-content/themes/core/patterns/card-post-search-result.php
```diff
@@ -14,13 +14,17 @@
 <!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|40","right":"0","bottom":"var:preset|spacing|40","left":"0"}}},"className":"p-card-search-result l-clearfix","layout":{"type":"default"}} -->
 <div class="wp-block-group p-card-search-result l-clearfix" style="padding-top:var(--wp--preset--spacing--40);padding-right:0;padding-bottom:var(--wp--preset--spacing--40);padding-left:0"><!-- wp:post-featured-image {"width":"22%","align":"right","className":"p-card-search-result__image s-aspect-ratio-cover s-aspect-ratio-4-3"} /-->
 
-<!-- wp:tribe/post-type-name {"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|10"}}}} /-->
+<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"tribe/post-type-name"}}},"className":"t-category","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|10"}}}} -->
+<p class="t-category" style="margin-bottom:var(--wp--preset--spacing--10)">Post Type Name Placeholder</p>
```
dpellenwood, line 18: I wonder if we should use realistic placeholder values and can we style the placeholder a little differently by default? Maybe a value like `Post` instead of `Post Type Name Placeholder` here, below, and in the example markup in the classes as well?

wp-content/themes/core/patterns/card-post-search-result.php
```diff
@@ -14,13 +14,17 @@
...
+<!-- /wp:paragraph -->
 
 <!-- wp:post-title {"style":{"spacing":{"margin":{"top":"0","bottom":"var:preset|spacing|10"}}},"className":"p-card-search-result__title","fontSize":"60"} /-->
 
 <!-- wp:post-excerpt /-->
 
-<!-- wp:tribe/post-permalink {"style":{"spacing":{"margin":{"top":"var:preset|spacing|20","right":"0","bottom":"0","left":"0"}}}} /-->
+<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"tribe/post-permalink"}}},"className":"t-caption","style":{"spacing":{"margin":{"top":"var:preset|spacing|20"}}}} -->
+<p class="t-caption" style="margin-top: var(--wp--preset--spacing--20)">Post Permalink Placeholder</p>
```
dpellenwood, line 26: Maybe a value like `http://example.com/hello-wold` styled differently?

wp-content/themes/core/templates/search.html
```diff
@@ -8,20 +8,26 @@ <h1 class="wp-block-heading has-text-align-center" style="margin-top:0;margin-bo
 <!-- wp:search {"label":"Search","showLabel":false,"buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true} /--></div>
 <!-- /wp:group -->
 
-<!-- wp:tribe/query-results-count {"align":"wide","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|30"}}}} /-->
+<!-- wp:paragraph {"align":"wide","metadata":{"bindings":{"content":{"source":"tribe/query-results-count"}}},"className":"tribe-query-results-count","style":{"spacing":{"margin":{"bottom":"var:preset|spacing|30"}}}} -->
+<p class="alignwide tribe-query-results-count" style="margin-bottom: var(--wp--preset--spacing--30)">Query Results Count Placeholder</p>
```
dpellenwood, line 12: Maybe a placeholder value like `99 results`?

dpellenwood (approved):
🟢  This is great Geoff. We'll review at the biweekly on Thursday and then we can merge. Thanks!

wp-content/themes/core/patterns/card-post-search-result.php
GeoffDusome: Do you have any ideas on styling the placeholders differently? Using bindings doesn't really give us anything to target to be able to style this differently. Only thing I was noticing was `contenteditable="false"` but not sure how reliable that is - the markup seemed different depending where the binding appeared in the template.

dpellenwood: Yeah, I wondered about that. Let's leave it as-is for now. Hopefully Core will provide us some additional classes/hooks that could be useful in the future.

dpellenwood:
@GeoffDusome you're good to merge this and send to QA when you have a minute.

PR #151, opened 2024-08-02
[CHORE] Github Actions/Lefthook updates

JMRhodes:
- **Chore:** Github Action workflow updates to use the latest versions of actions/cache.
- **Chore:** Github Action tests job removed to conserve resources. Tests should be run locally.
    - This is a somewhat costly job to run on a project that we aren't specifically writing tests for. I suggest we leave the workflow config to be enabled on a per-project basis if deemed necessary.
- **Updated:** Composer PHP minimum requirement set to 8.2 to match expected platform version.
- **Updated:** GHA Code quality workflow dispatch to Pull Requests opened.
- **Updated:** Lefthook commit script to include phpstan.


.github/workflows/code-quality.yml
```diff
@@ -28,10 +27,3 @@ jobs:
     uses: ./.github/workflows/static-analysis.yml
     secrets:
       COMPOSER_AUTH_JSON: ${{ secrets.COMPOSER_AUTH_JSON }}
-
-  slic:
```
dpellenwood, line 32: I wonder if we should comment this out rather than remove it entirely?

.github/workflows/code-quality.yml
```diff
@@ -28,10 +27,3 @@ jobs:
...
-    name: 'PHP Tests'
-    needs: [coding-standards, phpstan, linting]
-    uses: ./.github/workflows/php-tests.yml
-    secrets:
-      COMPOSER_AUTH_JSON: ${{ secrets.COMPOSER_AUTH_JSON }}
```
dpellenwood, line 37: I wonder if we should comment this out rather than remove it entirely?

.lando.yml
```diff
@@ -5,7 +5,7 @@ config:
   database: mariadb
   composer_version: 2-latest
   via: nginx
-  xdebug: true
+  xdebug: false
```
dpellenwood, line 8: Curious: do you prefer to start projects w/o xdebug enabled and only enable it when needed?

dpellenwood (approved):
Couple questions, but I'm fine with this. 🐐 💨

.lando.yml
JMRhodes: Yeah thinking more about FE devs here who typically don't use it. Depending on your local config, this can print out a lot of errors if enabled and not configured properly in your IDE. So this kind of avoids that by default.

.github/workflows/code-quality.yml
JMRhodes: I thought about that. I'm fine with doing that if we're ok committing commented code.

dpellenwood: In this case, I think it make sense as it may be used in the future.

JMRhodes: Updated

dpellenwood:
@JMRhodes you're good to merge this when you have a couple minutes.

PR #152, opened 2024-08-05
[MOOSE-99] Moose 2.0

GeoffDusome:
## What does this do/fix?

- **Blocks**: Removes unused block definitions from core plugin. Since we've moved to defining block variations in the `styles` directory, it's no longer necessary to define block variations in the PHP classes that extend `Block_Base`. 
- **Blocks**: I've moved all variation definitions to separate JSON files within the `/styles` directory; I'm on the fence here about restructuring this directory to additionally house the block styles that live within `theme.json` currently (to be something like `styles/blocks` & `styles/variations`). This would allow us to reduce the size of the `theme.json` file, but would split things up a bunch and possibly make finding things confusing. 
- **Color**: Removes empty color utilities file
- **Color**: Various color updates; Note here that I did _not_ update colors to use `primary`, `secondary`. I wasn't quite sure why design made this decision, but I'm 99% sure we wanted to get away from doing this. Please let me know if I'm just wrong here. 
- **Layout**: Breakpoints have been updated to match design.
- **General**: Removed transitions that aren't used in core Moose; I personally haven't used any of these easings and I've only used the general `--transition`. 
- **General**: Moves some files around to places where they make a bit more sense; `focus-visible` -> `/actions`; stacking order -> `columns` block.
- **General**: Icons have been reduced to only using one definition for each icon. We should be trying to use `mask` where possible so that we don't need to define multiple icon colors. 
- **General**: `scss` files in the theme should now more consistently use themed defined CSS properties, rather than WP preset properties.
- **Typography**: Added overline block style for the Paragraph block; Added matching utility class / mixin
- **Typography**: Added font selection styles; Do we need dark theme changes for this?
- **Typography**: Removes Roboto fonts; switches to Inter Tight. Font files gathered from [Google Webfonts Helper](https://gwfh.mranftl.com/fonts/inter-tight?subsets=latin) since there's still no reliable way to use Google Fonts in `theme.json`
- **Buttons**: Added "Has Icon" toggle to button block; Allows us to define when an arrow shows, instead of it being default for each button. This also allows us to have additional button styles without needing to define separate button styles for each.
- **Buttons**: You'll notice that there's theme styles in the Button block styles; Unfortunately this is a side-effect of the block variation JSON files not being able to parse variations within variations (as described in [this Slack thread](https://tribe.slack.com/archives/C01CY7418GY/p1722639036521929)). This is a temporary solution for now until this gets resolved in a later WP release.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-99)
- [Deployed PR on Dokku](http://moose-dev-pr152.d1.moderntribe.qa/)

Screenshots/video:
- TBD


wp-content/themes/core/assets/pcss/color/_variables.pcss
```diff
@@ -13,69 +13,65 @@
...
-	--color-neutral-90: var(--wp--preset--color--neutral-90);
+	--color-neutral-10: var(--wp--preset--color--neutral-10, #f5f5f5);
+	--color-neutral-20: #e8e8e8;
+	--color-neutral-30: #d1d1d1;
+	--color-neutral-40: #a3a3a3;
+	--color-neutral-50: var(--wp--preset--color--neutral-50, #767676);
+	--color-neutral-60: #5d5d5d;
+	--color-neutral-70: #464646;
+	--color-neutral-80: #2f2f2f;
+	--color-neutral-90: #1f1f1f;
```
dpellenwood, line 33: Curious about this change. IIRC, our goal was to always define all properties via theme.json and then reference those values in our CSS to prevent mismatched values.

wp-content/themes/core/assets/pcss/color/_variables.pcss
```diff
@@ -13,69 +13,65 @@
 	 * Color Naming Site: https://chir.ag/projects/name-that-color/
 	 * ----------------------------------------------------------------------- */
 
-	--color-royal-blue: var(--wp--preset--color--royal-blue);
-	--color-jacksons-purple: var(--wp--preset--color--jacksons-purple);
+	--color-blue: var(--wp--preset--color--blue, #3050e5);
+	--color-blue-rgb: 48, 80, 229;
```
dpellenwood, line 17: Couldn't the blue preset value be the rgb() color?

wp-content/themes/core/assets/pcss/media-queries/_variables.pcss
```diff
@@ -7,22 +7,26 @@
  *
  * ------------------------------------------------------------------------- */
 
-@custom-media --viewport-xxsmall (min-width: 320px);
-@custom-media --viewport-xsmall (min-width: 500px);
-@custom-media --viewport-small (min-width: 600px);
-@custom-media --viewport-small-max (max-width: 767px);
-@custom-media --viewport-medium (min-width: 768px);
-@custom-media --viewport-wpadmin (min-width: 783px);
```
dpellenwood, line 15: We should keep the wpadmin breakpoint defined (even if we're using it for tablet). Often designs don't stick to that value when they get implemented.

wp-content/themes/core/assets/pcss/typography/_mixins.pcss
```diff
@@ -112,10 +123,10 @@
...
 	color: var(--color-black);
-	font-family: var(--wp--preset--font-family--roboto);
-	font-size: var(--wp--preset--font-size--10);
+	font-family: var(--font-family-sans-serif);
+	font-size: var(--font-size-10);
 	line-height: 1.6;
-	font-weight: var(--wp--custom--font-weight--regular);
+	font-weight: var(--font-weight-regular);
 	text-transform: uppercase;
 	text-decoration: none !important; /* override block editor default */
```
dpellenwood, line 131: Do we still need this override?

wp-content/themes/core/assets/pcss/typography/_mixins.pcss
```diff
@@ -151,13 +162,13 @@
 	align-items: center;
 	padding: 4px 16px;
 	background-color: var(--color-neutral-20);
-	border-radius: 100px;
+	border-radius: var(--border-radius-full);
 	color: var(--color-black);
 	text-decoration: none !important; /* override block editor default */
```
dpellenwood, line 167: And this override?

wp-content/themes/core/styles/blocks/heading/b-x-small.json
```diff
@@ -0,0 +1,13 @@
+{
```
dpellenwood, line 1: Rather than naming these with letters, can we use a 10-based index system? Ie.  `10-xx-small.json`, `20-x-small.json` etc? using 10-based indexes allows for adding new variations between existing ones easier.

dpellenwood (commented):
Left some questions & suggestions, Geoff. This is great. We'll review at the biweekly together to get feedback before we approve to merge.

wp-content/themes/core/assets/pcss/color/_variables.pcss
GeoffDusome: The issue I've been running into on projects lately is that design wants all the colors available, but they don't want all the colors available to editors. So, we end up only adding a few of the colors to the global color palette and then the rest get added in code to be used elsewhere on the site. Normally these other grays are used in places not editable by editors like buttons and navigation. It's likely though that we're not using all of these colors and a lot of them can be stripped though. 

Design also wanted to go back to `primary` / `secondary` with this update - so maybe we have a pretty big disconnect on colors that we need to meet to go over? Maybe I can post in the channel first for feelers on if we need to meet or not?

There's always the option to get super granular with colors too, where we can choose per block which colors are available in the palette.

wp-content/themes/core/assets/pcss/color/_variables.pcss
GeoffDusome: I wasn't able to find a combination of this that worked. It doesn't appear you can just assign `48, 80, 229` to the preset value. It expects a prefix like `#` or `rgb(`. Setting it to `rgb(48, 80, 229)` doesn't allow you to use in an `rgba()` definition. 

Since this color is only used for the `::selection` code, I could also just one-off the `rgba()` definition to the selection code instead?

wp-content/themes/core/assets/pcss/media-queries/_variables.pcss
GeoffDusome: I removed this because it no longer is true and now aligns with the `--mq-wp-medium` media query at the bottom of this file. Maybe the breakpoints have been updated in core, but `783px` is no longer mentioned in here: https://github.com/WordPress/gutenberg/blob/trunk/packages/base-styles/_breakpoints.scss.

I could rename `medium` to be `wpadmin` or `adminbar` so it's more in line with what it's used for generally?

wp-content/themes/core/assets/pcss/typography/_mixins.pcss
GeoffDusome: I believe this is used in the Post Terms block to style category / tag which uses links to display the categories / tags. I believe by default these links use underline, hence the override. Same applies to the override below.

wp-content/themes/core/assets/pcss/media-queries/_variables.pcss
dpellenwood: Interesting. Let's go with `adminbar`. We often use a different medium value on projects.

wp-content/themes/core/assets/pcss/color/_variables.pcss
dpellenwood: Let's keep it defined as a color var, but maybe name it specifically like `--color-selection-highlight` or something similar?

dpellenwood:
@GeoffDusome , left a couple follow-ups to your questions. Feel free to merge after you've reviewed/updated. Thanks!

wp-content/themes/core/assets/pcss/color/_variables.pcss
GeoffDusome: Updated to new naming for variable; Moved it to it's own section in the colors file; Added a description for this new section and updated the value to use a #RRGGBBAA format.

wp-content/themes/core/assets/pcss/media-queries/_variables.pcss
GeoffDusome: Updated this variable to be `--mq-wp-adminbar`. This breakpoint will match the defined `tablet` media query but can be used in places where contexts dictates it makes more sense to reference the admin bar media query rather than the tablet media query.

dpellenwood (approved):
🥔 💨

PR #156, opened 2024-08-26
[MOOSE-99] Moose 2.0 DQA

GeoffDusome:
## What does this do/fix?

- updates highlight color to work with dark themes as well
- fixes issue with image block in editor while using grid width where the image couldn't span the entire width of the container
- fixes pagination arrows being invisible after removing some of the icons used for pagination; some colors were also not updated in this file so those were updated as well.

**Note: There was a bunch of DQA feedback here that was a result of WP 6.6 updating how the the block nesting padding works on Group blocks set to align full. Just _another_ reminder that some learnings that we may have had previously with layout now have been addressed in core and we may have to unlearn some tricks we used to bypass these issues.**

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-99?focusedCommentId=164747)


dpellenwood (approved):
🐳 💨

PR #161, opened 2024-11-05
BE: Remove SquareOne libs dependencies

MlKilderkin:
## What does this do/fix?

Remove SquareOne libs dependencies
Add required abstract classes
Update tests

## QA

Links to relevant issues
- https://moderntribe.atlassian.net/browse/MOOSE-168


dpellenwood (approved):
Looks great, @MlKilderkin ! I'd like @JMRhodes to give this one a look as well before you merge.

wp-content/plugins/core/src/Object_Meta/Meta_Map.php
```diff
@@ -0,0 +1,56 @@
+<?php declare(strict_types=1);
+
+namespace Tribe\Plugin\Object_Meta;
+
+/**
+ * Class Meta_Map
+ *
+ * Maps requests for meta keys to the Meta_Group responsible for handling it
+ */
+class Meta_Map {
```
JMRhodes, line 10: Did these come over from Libs? Wondering if that's still necessary. Similar question with `Meta_Group` and `Meta_Respository`?

MlKilderkin: Yes @JMRhodes this is from Libs. If we don't need something like this
```
public function get_meta( string $key ): mixed {
		return $this->meta->get_value( $this->post_id, $key );
	}
``` 
for meta pulling we can remove this

JMRhodes: @MlKilderkin what are your thoughts on it? Do you find it useful in your projects? Certainly fine keeping it in there if we're utilizing it.

MlKilderkin: Sorry, I totally forgot about this. 

@JMRhodes Personally I prefer to use get_post_meta, specially if I have some custom places where I can't(or don't need) to inject Meta dependency in order to get only 1 item. Since we don't create complex components like on S1 it can be removed

MlKilderkin:
@dpellenwood @JMRhodes I removed Meta classes as discussed from above. However, looks like I missed required classes for post registration. After latest changes were merged I found missing items and added them as well. Please take a look

JMRhodes (approved):
🟢 Thanks for tackling this!

PR #163, opened 2024-11-15
Adds required PHP extensions so they get added to the Dokku envs.

dpellenwood:
## What does this do/fix?

Our Dokku dev sites expect any PHP extensions to be stated as composer requirements. This PR adds missing extensions that [WordPress expects](http://p.tri.be/i/ckegk3) so they are properly configured on Dokku. Note that we're skipping `imagick` for now as it has [some conflicts for Docker-based PHP images such as Lando](https://github.com/lando/php/issues/114) and is not available in those images at this time. This is acceptable because only GD or Imagick are required for image processing, not both.


JMRhodes (approved):
🟢

PR #164, opened 2024-11-19
Feature: 1Password CLI Integration

dpellenwood:
## What does this do/fix?
This work swaps our our manual process of creating composer auth.json files via copy-paste and the json object blob in the `composer_auth_json` GitHub secret for an automated process via the 1Password CLI. This automation works for both local dev and GitHub actions.

For local dev, I've created a composer script generates the auth.json file from the template using the local developer's 1Pass account. (This requires the [1Password CLI](https://developer.1password.com/docs/cli/get-started) to be installed and configured for a user first.)

For GH Actions, a specific [1Pass Service Account](https://developer.1password.com/docs/service-accounts/) is used to authenticate against the required vault.

This PR also updates our documentation in GitHub to be more robust and concise. 

To do:
- [ ] Create a demo video of this feature
- [x] Define & test env-specific secrets for GH Actions
  Confirmed that setting an environmental secret to override the repository secrets works. It requires that the workflow jobs utilizing the env-specific secrets define an `environment` that matches the Environment named in the repository configuration.
- [x] Update Readme docs with the new project setup process
- [x] Write up complete documentation on the 1Password integration 
- [x] ~Update deploy actions with the new composer install action~
- [x] How best to handle 1Password service accounts for GHA integration.
  - 1Password Service Accounts can't be modified and can't add new vault access automatically, so a new Service account for each vault that needs to be accessed should be created.
- [x] How best to handle different license key values from clients.
  1. Create a new 1Pass item in the project's vault using the MT-Composer-Auth item as a template.
  2. Create a new 1Password Service Account with read access to the project's vault.
  3. Create project-specific GitHub secrets to hold the correct 1Pass Service Account Token (`OP_SERVICE_ACCOUNT_TOKEN`), vault (`OP_VAULT`), and item (`OP_ITEM`).
  4. Future Improvement: One thing I haven't examined closely is how we might use different secrets per environment with the same projects such that deploying to Dokku can use the MT vault items and the client's env's can use their own vault items.
- [x] How best to handle additions or changes to the `auth.template.json file`. For example, how to add `github-oauth` token and other new auth values (Yoast SEO, etc) with the least amount of friction. 
  1. Add a new section and item(s) to the specific project's 1pass item
  2. Update the `auth.template.json` file with the path to the new section & item(s)
- [x] ~Ideally: How might we automate the script run for local dev. This needs to run on the local machine, outside of the Lando containers, as it requires the user's 1Pass account be setup to auth against the vault.~
  - There is no good way to do this due to how Lando is designed to run all actions within the containers it spawns. I've modified the project setup instructions such that the first command to run is `composer create-auth`, before running `lando start`.


.github/workflows/phpcs.yml
```diff
@@ -48,10 +51,21 @@ jobs:
...
+      - name: Install 1Password CLI
+        if: steps.changed-files.outputs.phpcs == 'true'
+        uses: 1password/install-cli-action@v1
+
+      - name: Create auth.json from 1Password
+        if: steps.changed-files.outputs.phpcs == 'true'
+        env:
+          OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
+        run: |
+          op user get --me
```
dpellenwood, line 63: Q: Do we want to output the service account user for each run? It's somewhat helpful, but also noisy. Thoughts?

.github/workflows/phpcs.yml
```diff
@@ -48,10 +51,21 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-composer-
 
+      - name: Install 1Password CLI
```
dpellenwood, line 54: Q: Should this action happen here or earlier in the job?

dev/scripts/create-auth-json.sh
```diff
@@ -0,0 +1,21 @@
...
+if ! command -v op > /dev/null 2>&1; then
+  echo >&2 "It appears that the 1Password CLI is not installed. Skipping.";
+  exit 0;
+fi
+
+# Exit if there is an existing auth.json file
+if [ -f auth.json ]; then
+  echo >&2 "auth.json already exists. Skipping.";
+  exit 0;
+fi
```
dpellenwood, line 15: To test: do we even need this? the inject command may already support this?

.github/actions/composer-install/action.yml
```diff
@@ -0,0 +1,56 @@
...
+      run: |
+        echo "OP_VAULT=${{ inputs.OP_VAULT || 'Engineering' }}" >> $GITHUB_ENV
+        echo "OP_ITEM=${{ inputs.OP_ITEM || 'MT-Composer-Auth' }}" >> $GITHUB_ENV
+
+    - name: Create auth.json via 1Password CLI
+      shell: bash
+      env:
+        OP_SERVICE_ACCOUNT_TOKEN: ${{ inputs.OP_SERVICE_ACCOUNT_TOKEN }}
+      run: |
+        op user get --me
```
dpellenwood, line 50: Question: Should I leave this command before the `op inject...` in each instance? [It's output](https://github.com/moderntribe/moose/actions/runs/11943174941/job/33291712843#step:4:83) can be helpful for debugging to see who the authenticated 1Password user is, but it may also be more info than needed and "leak" sensitive info we don't need.

.github/actions/composer-install/action.yml
```diff
@@ -0,0 +1,56 @@
+name: Composer Install
+
+inputs:
+  OP_SERVICE_ACCOUNT_TOKEN:
+    required: true
+    description: 1Password service account token to use for populating composer auth.json
+  OP_VAULT:
```
dpellenwood, line 7: Thoughts on the name of these secrets? Should the be more verbose?

crstauf: > Secret names can only contain alphanumeric characters ([a-z], [A-Z], [0-9]) or underscores (_). Spaces are not allowed. Must start with a letter ([a-z], [A-Z]) or underscores (_).

😞

.github/workflows/deploy-dev.yml
```diff
@@ -14,41 +14,21 @@ jobs:
...
-          php-version: '8.2'
-        env:
-          COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
       # Get Build Repository
       - name: Check out build branch
         uses: actions/checkout@v4
         with:
           path: ${{ env.BUILD_FOLDER }}
           ref: ${{github.ref}}
```
marcossantiago, line 23: This is an old deploy to WPEngine and it has been deprecated. This should be removed. Samve for staging and Production.

.github/actions/composer-install/action.yml
crstauf: `ONEPASS` would be more clear than `OP`.

marcossantiago (approved):
minor comment regarding deprecated github actions workflows

.github/workflows/deploy-dev.yml
dpellenwood: I'm pretty sure projects still use these deploy workflows after cloning from Moose. I'll check w/ the BE team and confirm though. Thanks, Marcos!

LayaTaal:
Going to review this for real @dpellenwood, but I'm really curious to know what "HitHub" is. Is that a place where assassins go to find their targets? 😆 

image

.github/actions/composer-install/action.yml
LayaTaal: I'm fine either way. I think `ONEPASS` is more clear as well, but we do have a description right below that should make `OP` clear. So either way is fine by me.

dpellenwood:
Fixed!

> Going to review this for real @dpellenwood, but I'm really curious to know what "HitHub" is. Is that a place where assassins go to find their targets? 😆
> 
> image

.github/workflows/code-quality.yml
```diff
@@ -27,12 +29,16 @@ jobs:
     needs: [coding-standards, linting]
     uses: ./.github/workflows/static-analysis.yml
     secrets:
-      COMPOSER_AUTH_JSON: ${{ secrets.COMPOSER_AUTH_JSON }}
+      OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
+      OP_VAULT: ${{ secrets.OP_VAULT }}
+      OP_ITEM: ${{ secrets.OP_ITEM }}
 
 # Enable this job if you have PHPUnit tests
```
LayaTaal, line 36: Out of curiosity, is it common to have PHP unit tests in Moose projects? Just wondering if this is something we maybe just want to remove (or move to Moose Resources). Not necessary in this PR, but wanted to note it.

LayaTaal (approved):
This is awesome @dpellenwood! I love all the updates to the docs here that you did along with the 1pass integration. Thanks for going through this. I'm curious to see how the automated auth.json setup will feel when we start using it.

.github/workflows/code-quality.yml
dpellenwood: Yeah, it's definitely worth a conversation about removing for now.

.github/actions/composer-install/action.yml
dpellenwood: After testing more, the 1Password CLI [assumes an env secret](https://developer.1password.com/docs/ci-cd/github-actions/#get-started) called `OP_SERVICE_ACCOUNT_TOKEN` to authenticate when using their GH Action. Seeing as this is their existing syntax and naming schema, I think it the most sense to stick to the same convention for our our own naming.

PR #166, opened 2024-11-22
Chore/moose 159/package updates

dpellenwood:
## What does this do/fix?
* Fixes a small typo with our composer post-install script.
* Updates WP Core to 6.7.1
* Adds Gravity Forms as a composer dep 💥 
* Updates plugins to latest
* Updates nvm to use latest LTS version of node (22.11.0)
* Updates NPM packages to latest versions
* Updates our browser support to make sure we're not excluding highly-used versions of Safari.

## QA

Links to relevant issues
- [MOOSE-159](https://moderntribe.atlassian.net/browse/MOOSE-159)

[MOOSE-159]: https://moderntribe.atlassian.net/browse/MOOSE-159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -11,9 +11,9 @@
 		"email": "admin@tri.be"
 	},
 	"browserslist": [
-		"last 2 versions",
```
dpellenwood, line 14: This was [excluding all versions of Safari under v18.0](https://browsersl.ist/#q=last+2+versions%2C%0Anot+op_mini+all%2C%0Anot+dead), which is bad as not many people have migrated to the latest version of MacOS/iOS yet.

package.json
```diff
@@ -11,9 +11,9 @@
...
-		"not op_mini all",
-		"not dead"
+		"defaults",
```
dpellenwood, line 14: Simpler [set of browsers](https://browsersl.ist/#q=defaults%2C%0Anot+ios+%3C+17.5&region=alt-na) to remember while still supporting the older versions of Safari we need.

GeoffDusome (approved):
🙌

package.json
```diff
@@ -11,9 +11,9 @@
...
+		"not ios < 17.5",
```
LayaTaal, line 15: Just curious why you decided to include this? It looks like without it we are included some 16.x versions of Safari? I just wonder if it is ok to keep those so we don't have to remember to keep updating this line as things move forward.

dpellenwood: Good question. Correct. The "defaults" key still supports 16.6, 16.7, and 15.6-8. We don't want or need to support those really old versions of Safari as they often lack features we want to use.

The usage on those versions is minimal; and, since those versions are [no longer supported by Apple](https://en.wikipedia.org/wiki/Safari_(web_browser)#Version_compatibility), I don't see a reason to keep them.

PR #171, opened 2024-12-16
[MOOSE-175]: Disable Font Library UI

GeoffDusome:
## What does this do/fix?

- Design has determined that the font library poses a risk to our Moose projects where it's possible that a client could break the design system by adding new font libraries or disabling existing ones. This is too dangerous of a feature to keep in Moose, so this PR disables this feature. 
- I've created a new `Theme_Config` class called `Block_Editor` to house configs like these. 

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-175)

Screenshots/video:
- Font library visible (prior to this change): http://p.tri.be/i/s3iEPn
- Font library removed: http://p.tri.be/i/AKqCEH


dpellenwood (changes requested):
One thought on where to handle this filter.

dpellenwood (approved):
🟢

PR #172, opened 2025-01-02
[MOOSE-183] Global Padding Issues

GeoffDusome:
## What does this do/fix?

- How WP handles global padding doesn't work well with how we're utilizing nested `.alignfull` elements to create layouts with the Cover block (the Hero Split pattern is a outlier here, but falls into a similar issue). 
- I went ahead and enabled background images for the Group block. Using this feature, rather than the Cover block, should allow us to create the layouts we're looking for without having to deal with nested `.alignfull` blocks. I went ahead and created an example and it looks pretty good, no padding issues! This solution also allows us to use a theme on the top level Group block of the pattern without having to worry about nested Group blocks problem we would have in the past. 
- As for the Hero Split pattern - my suggestion here would just be to use the default padding controls to add left/right padding. This is a one-off issue now that is fixable without adding any code. 
- Once this PR goes through, I'll go through the patterns on the Moose dev environment and update them to use the Group block background image, rather than the Cover block.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-183)

Screenshots/video:
- Example Group block background image section: http://p.tri.be/i/SkGbEV


dpellenwood (approved):
Sounds good to me! just one ask: add a changelog entry? I've somehow neglected updating the change log for months. I'll go back through and update for all the PRs we've merged since the last entry in a separate PR.

PR #173, opened 2025-01-02
[MOOSE-181] Table Overflow

GeoffDusome:
## What does this do/fix?

- adds content width as a layout variable
- sets `min-width` on table elements within Table blocks. As the comment states, this is the only way to actually apply the `overflow-x` set on the Table block for tables that have minimal content. 
- I'll put this one through DQA so designers can confirm what minimum width they'd like to see.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-181)

Screenshots/video:
- Mobile table view: http://p.tri.be/i/AzzZSr


dpellenwood (approved):
🟢  Add a changelog entry, please!

PR #176, opened 2025-01-03
Lando Project Setup Improvements and Node integration

dpellenwood:
## What does this do/fix?

* Adds a bash script for managing the installed version of WordPress via Composer. This make it simpler to manage the WP version constant and prevents the CLI download/install command from running unless the version of WP has changed in composer.json.
* Add a PHP script to automatically generate both local-config files (PHP & JSON) and properly populate local-config.json with values from Lando's project configuration.
* Adds a Node service to the Lando config so that the npm packages can be automatically installed and the theme assets automatically built each time the Lando project starts. Note that this isn't intended to prevent using `nvm`, installing the nom packages locally or running the npm scripts locally for frontend development. It's only intended to ease onboarding to a new project for developers who don't need to worry about regularly installing the npm packages or building the assets.  
* Misc documentation updated for the changes above.

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -33,6 +33,11 @@ services:
 #      - appserver_nginx
 #      - appserver
 
+events:
+  post-start:
+    - appserver: composer install
+    - node: npm install && npm run build
+
 tooling:
   xdebug-on:
     service: appserver
--- a/.lando.yml
+++ b/.lando.yml
@@ -1,5 +1,6 @@
 name: moose
 recipe: wordpress
+
 config:
   php: '8.3'
   database: mariadb:11.5
@@ -7,11 +8,11 @@ config:
   via: nginx
   xdebug: false
   memcached: true
+
 services:
   appserver:
-    run:
-      - composer install
-      - composer run copy-local-configs
+    build:
+      - composer create-local-configs
     overrides:
       environment:
         - XDEBUG_TRIGGER=1
--- a/.lando.yml
+++ b/.lando.yml
@@ -16,6 +16,8 @@ services:
     overrides:
       environment:
         - XDEBUG_TRIGGER=1
+  node:
+    type: node:22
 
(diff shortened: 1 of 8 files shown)


.lando.yml
```diff
@@ -30,7 +33,14 @@ services:
 #      - appserver_nginx
 #      - appserver
 
+events:
+  post-start:
+    - appserver: composer install
+    - node: npm install && npm run build
```
dpellenwood, line 39: Thoughts on automatically running these each time the project is started? I see it as a convenience to not have to remember to do these steps in the case that a composer or npm dependency has changed or the theme assets need to be rebuild. However, it does require more time to perform these actions on `lando start`.

Vinsanity: @dpellenwood Is there any way to preset a prompt for this? `Do you wish to build FE assets? [Y/n]`

dev/scripts/create-local-configs.php
```diff
@@ -0,0 +1,59 @@
+<?php declare(strict_types=1);
```
Vinsanity, line 1: ❤️

LayaTaal (approved):
This is sweet! Nice improvements to make things move more quickly. Thanks Dave!

.lando.yml
```diff
@@ -1,20 +1,23 @@
...
-    run:
-      - composer install
-      - composer run copy-local-configs
+    build:
+      - composer create-local-configs
     overrides:
       environment:
         - XDEBUG_TRIGGER=1
+  node:
+    type: node:22
```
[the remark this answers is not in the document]
dpellenwood: Unfortunately, no. It has to be a numbered release. :(

.lando.yml
lindseyolson: I'm okay with this each time too

lindseyolson (approved):

PR #177, opened 2025-01-06
[MOOSE-178]: Prevent 404 on any link if not authorized

MlKilderkin:
## What does this do/fix?

[MOOSE-178]: Prevent 404 on any link if not authorized
Training CPT will be hidden by default because of [this](https://github.com/moderntribe/moose/blob/main/wp-content/plugins/core/src/Post_Types/Training/Training_Subscriber.php#L33)

## QA

Links to relevant issues
-  https://moderntribe.atlassian.net/browse/MOOSE-178
- 

[MOOSE-178]: https://moderntribe.atlassian.net/browse/MOOSE-178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MlKilderkin:
I am actually not even sure that we need this [hook](https://github.com/moderntribe/moose/blob/main/wp-content/plugins/core/src/Post_Types/Training/Training_Subscriber.php#L37)
Using [hook](https://github.com/moderntribe/moose/blob/main/wp-content/plugins/core/src/Post_Types/Training/Training_Subscriber.php#L32) here always returns false for non-authorized users. This prevents setting of the CPT query variable in the core code. As the result, if you go to the CPT link you will see the default loop page

crstauf:
> This prevents setting of the CPT query variable in the core code. As the result, if you go to the CPT link you will see the default loop page

Exactly why the hook was added: natural behavior is to 404 for pages that don't exist.

dpellenwood (approved):
🟢

MlKilderkin:
> > This prevents setting of the CPT query variable in the core code. As the result, if you go to the CPT link you will see the default loop page
> 
> Exactly why the hook was added: natural behavior is to 404 for pages that don't exist.

My concern there that `template_redirect` doesn't work as 404 in such case. When I debug this locally I see next picture:
1. `is_post_type_viewable` works and return false
2.  query variable is not set for Training CPT because of `is_post_type_viewable`
3. `template_redirect` sees `post` post type and shows default loop page for Training slug

We might need to change this to the `pre_handle_404` even though it works after `template_redirect`

crstauf:
> My concern there that `template_redirect` doesn't work as 404 in such case.

Works for me:

2025-01-06T182817

dpellenwood:
@MlKilderkin curious if you're still blocked on this PR of it can be merged and sent for QA?

MlKilderkin:
> @MlKilderkin curious if you're still blocked on this PR of it can be merged and sent for QA?

I think we can go ahead and merge it

PR #178, opened 2025-01-06
[MOOSE-181] Table Overflow / Scrolling DQA

GeoffDusome:
## What does this do/fix?

Design is requiring that the table block be offset (to make it more clear that it can be scrolled on mobile). Unfortunately, it would be almost impossible to handle all situations for this because of different block widths / alignments and just general over-abundance of use-cases for the table block.

In order to implement, we need to consider different block widths, as well as overriding the default max-width of some of these blocks. This also forces us to choose a breakpoint at which this code gets implemented. I've chosen a tablet size. 

I am aware that this is a very messy solution and I'm hoping upon seeing this that maybe we can talk design out of this as it adds so much unnecessary bloat to the simple solution we added previously (just adding the min-width to the table element). I think the main point here is that there are so many use-cases for the table element that we can't just implement a grid offset for every case. 

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-181?focusedCommentId=168191)

Screenshots/video:
- Offset / overflowed table block: http://p.tri.be/i/ObHju3


GeoffDusome:
@dpellenwood I've tagged you in [the ticket here](https://moderntribe.atlassian.net/browse/MOOSE-181?focusedCommentId=168192). I'll move the ticket back into DQA with a note for them to chat with you. Closing this PR.

PR #180, opened 2025-01-13
[MOOSE-172] Admin Menu Cleanup

GeoffDusome:
## What does this do/fix?

Please note I do not know if this is the best way to go about this. The only other way I could think of would be to awkwardly handle reordering all the menu item priorities, which would get super messy IMO. 

This PR assumes that newly created CPTs would need to be manually added to the menu order array in the correct spot.

- reorder menu items via `menu_order` filter. 
- update Training Docs menu item label

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-172)

Screenshots/video:
- Updated menu order: http://p.tri.be/i/42Jcpi

### DIFF
--- a/wp-content/plugins/core/src/Core.php
+++ b/wp-content/plugins/core/src/Core.php
@@ -29,6 +29,7 @@ class Core {
 		Integrations\Integrations_Subscriber::class,
 		Object_Meta\Meta_Subscriber::class,
 		Theme_Config\Theme_Config_Subscriber::class,
+		Menus\Menu_Subscriber::class,
 
 		// Post Types
 		Post_Types\Page\Page_Subscriber::class,
--- a/wp-content/plugins/core/src/Core.php
+++ b/wp-content/plugins/core/src/Core.php
@@ -23,13 +23,13 @@ class Core {
 	 * @var string[] Names of classes extending Abstract_Subscriber.
 	 */
 	private array $subscribers = [
-		Settings\Settings_Subscriber::class,
 		Assets\Assets_Subscriber::class,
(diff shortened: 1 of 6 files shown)


crstauf (approved):

wp-content/plugins/core/src/Menus/Menu_Order.php
```diff
@@ -0,0 +1,39 @@
+<?php declare(strict_types=1);
+
+namespace Tribe\Plugin\Menus;
+
+class Menu_Order {
```
dpellenwood, line 5: I would make sure to prefix your work with `Admin_` as this `menu` namespace is often used for frontend menus as well.

wp-content/plugins/core/src/Core.php
```diff
@@ -29,6 +29,7 @@ class Core {
 		Integrations\Integrations_Subscriber::class,
 		Object_Meta\Meta_Subscriber::class,
 		Theme_Config\Theme_Config_Subscriber::class,
+		Menus\Menu_Subscriber::class,
```
dpellenwood, line 32: Minor, but it can be nice to keep these in alpha order.

dpellenwood (changes requested):
This is a nice start, @GeoffDusome . Some changes & questions. I'd like @JMRhodes to weigh in on this one, too, please!

GeoffDusome:
@dpellenwood Solved all, please take another look when you can.

JMRhodes (approved):
🟢

dpellenwood (approved):
🟢

PR #181, opened 2025-01-13
[MOOSE-179]: add details block summary link styling

GeoffDusome:
## What does this do/fix?

- style links within Details block `summary` elements.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-179)

Screenshots/video:
- Updated link styling: http://p.tri.be/i/CRb0tb

### DIFF
--- a/wp-content/themes/core/blocks/core/details/style.pcss
+++ b/wp-content/themes/core/blocks/core/details/style.pcss
@@ -1,0 +1,21 @@
+/**
+ * Styles specific to this block
+ */
+
+.wp-block-details {
+
+	/**
+	 * Summary elements shouldn't contain text markup, but for some
+	 * reason WP allows this markup within the summary text if you
+	 * copy & paste it into the summary element. We should try to
+	 * avoid this pattern, but since it's possible, we need to support
+	 * it. This only applies to links within the summary element.
+	 */
+	summary {
+
+		a {
+
+			@mixin inline-text-link;
+		}
+	}
+}
--- a/wp-content/plugins/core/src/Blocks/Blocks_Definer.php
+++ b/wp-content/plugins/core/src/Blocks/Blocks_Definer.php
@@ -11,6 +11,7 @@ use Tribe\Theme\bindings\Query_Results_Count;
 use Tribe\Theme\blocks\core\button\Button;
 use Tribe\Theme\blocks\core\column\Column;
 use Tribe\Theme\blocks\core\columns\Columns;
+use Tribe\Theme\blocks\core\details\Details;
 use Tribe\Theme\blocks\core\embed\Embed;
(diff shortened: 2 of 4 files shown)


wp-content/themes/core/blocks/core/details/style.pcss
```diff
@@ -0,0 +1,21 @@
...
+	 * reason WP allows this markup within the summary text if you
+	 * copy & paste it into the summary element. We should try to
+	 * avoid this pattern, but since it's possible, we need to support
+	 * it. This only applies to links within the summary element.
+	 */
+	summary {
+
+		a {
+
+			@mixin inline-text-link;
```
dpellenwood, line 18: Curious could we just add summary to the list of parents [here](https://github.com/moderntribe/moose/blob/main/wp-content/themes/core/assets/pcss/typography/anchors.pcss#L20-L23)? Or, even better, can we look at removing the specific parents for links in general? Has the editor improved enough to allow that?

dpellenwood (commented):
Question about this one, if you'd be so kind.

wp-content/themes/core/blocks/core/details/style.pcss
GeoffDusome: To your first question, yes, we absolutely could. Is that what you would want to see rather than the block specific styling?

To your second (& third) questions. I don't think we can do this while using the mixin. Targeting "`a`" globally is super dangerous for us in Gutenberg as there are so many elements that are links that aren't the inline links we're wanting to target. Of course, we do have the ability to target `link` elements within specific blocks using the `elements` > `link` hierarchy in `theme.json`, but without support for a lot of CSS properties, we'd be looking at adding a `css` attribute in `theme.json`, which I absolutely hate the idea of (we had to do this for the `spacer` block variations and I'm super not happy with that implementation).

dpellenwood: > Is that what you would want to see rather than the block specific styling?
I think so, yes. Otherwise, we're applying the same styling two completely different ways.

>  I don't think we can do this while using the mixin.
I figured as much, yeah. I do wonder if we're too restrictive with our parent selectors on that default anchors code. We should probably include headings, and other text elements that can contain links there, too. or maybe reverse it and use those styles for _all_ links that don't have a css class of any kind?

GeoffDusome: @dpellenwood Updated this - I also removed the additional `focus-visible` selector as the `inline-text-link` mixin already applies that to the same elements as the new selector we're targeting. [Here's a couple different blocks with links applied](http://p.tri.be/i/ynZSL1) (within the content section, we don't apply this same thing to the footer which is why there's still that blue text there).

PR #183, opened 2025-01-17
Feature/moose 157/theme configuration updates

Vinsanity:
## What does this do/fix?

Global Editor style and configuration changes.

- Added support for `blockGap` settings in specific editor blocks.
  - Columns, Group, and Navigation
- Fix the way that the editor stylesheet was being loaded in the editor.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-157)

Screenshots/video:
- [Link to Video](https://www.loom.com/share/66225c2bd3d64d258b21ba120f05a104?sid=f7c1147d-4e07-42c8-a27d-dab20f7e08af)


GeoffDusome (approved):

wp-content/plugins/core/src/Assets/Editor_Assets_Enqueuer.php
```diff
@@ -11,7 +11,13 @@ class Editor_Assets_Enqueuer extends Assets_Enqueuer {
 	public function register(): void {
 		$args = $this->get_asset_file_args( $this->assets_path . self::ASSETS_FILE );
 
-		add_editor_style( $this->assets_path_uri . self::EDITOR_FILE_NAME . '.css', );
+		wp_enqueue_style(
```
dpellenwood, line 14: I think you need to add an `is_admin()` check at the top of this function and bail early if the current view is not the admin.

dpellenwood (changes requested):
One change to make sure the editor scripts & styles don't end up bleeding to the frontend.

wp-content/plugins/core/src/Assets/Editor_Assets_Enqueuer.php
Vinsanity: Ah good call @dpellenwood!  this one is updated, tested, and fixed now.

GeoffDusome:
@Vinsanity @dpellenwood I think this needs to change again now that WP 6.8 added a new way to enqueue block styling. Maybe a separate PR later on though.

Vinsanity:
> @Vinsanity @dpellenwood I think the way we enqueue block styles needs to change again now that WP 6.8 added a new way to enqueue block styling. Maybe a separate PR later on though.

lol of course they changed it again.

dpellenwood (approved):
🍏

PR #188, opened 2025-03-12
[MOOSE-193] Create WP Controls Script Fixes

GeoffDusome:
## What does this do/fix?

- Updates Column block `stackingOrder` attribute to be a `number` field to prevent type mismatching when checking / applying the value
- When using `toggle` control types, you previously weren't able to set the `defaultValue` to `true` as there were incorrect checks in place. These checks have been updated to ensure that the `number` and `select` control types continue to function, and the toggle control can now function as it was supposed to. We should now be properly checking if the `toggle` control is set to `true` before applying the class / style rules, instead of just applying them if the value was different than the default. 

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-193)


dpellenwood (approved):
🟢

PR #194, opened 2025-05-01
Feature/moose 156/masthead

Vinsanity:
### Masthead and Mega Navigation for Moose!

**Please check out the demo video below. use 1.5-2x speed as it is 13min long.**

Adds several new tribe blocks (not acf)
- Menu Toggle for mobile menu toggling
- Site Search for masthead search overlay
- Mega Navigation Menu Item for creating mega menu content
- Mega Navigation Standard Menu Item for creating standard links or dropdown menus
- Navigation Link for creating a simple link and it utilizes the LinkControl component. This is to easily search for published singles within the site or custom links

Adds new template patterns and parts
- Adds a new pattern for mega menu content that is based off the footer pattern for enhanced navigation
- Adds a new template part for `mega-nav` for easier editing of mega nav content
- Adds a new template part for `utility-nav` for easier editing of utility nav content

I expect a lot of feedback here and I'm open to any suggestions for naming(feels confusing right now), for structure, consolidation/pruning of code, etc.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-156)

Screenshots/video:
- [Link to Video](https://www.loom.com/share/669f02117e894c97a4d487d3db9c97b1?sid=595b36ff-6018-4d37-9e6e-5331c62bf3f5)


wp-content/themes/core/blocks/tribe/site-search/block.json
```diff
@@ -0,0 +1,19 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/block.json",
```
LayaTaal, line 2: How useable is this block in other areas, like on pages, etc.? I'm wondering if this should be limited to being used in the masthead or not.

wp-content/themes/core/blocks/tribe/site-search/render.php
```diff
@@ -0,0 +1,28 @@
+<?php declare(strict_types=1);
+
+$search_icon = '<svg width="32" height="33" viewBox="0 0 32 33" fill="none" xmlns="http://www.w3.org/2000/svg">
```
LayaTaal, line 3: Can we store the icon as an asset and load it from the url?

LayaTaal (commented):
@Vinsanity I think this is an excellent first version of this. It works great out of the box and has almost all of the "must haves" IMO. I haven't gone through the complete code yet but wanted to drop my notes thus far. This is mostly from a user standpoint.

### Need to have
- [ ] Top level menu items need `cursor: pointer`
- [ ] Utility menu items need `cursor: pointer`
- [ ] Utility menu top level items don't have a hover state

### Nice to have
- [ ] I think it would be nice if clicking outside of a dropdown also closes the menu (like the escape key does)
- [ ] On mobile, when collapsing a menu the focus stays on that menu item. The blue text and heavy underline feel confusing since it is collapsed 
- [ ] On mobile, if I have opened a menu and then open another it collapses the previous one. This can feel a little jarring to me as it shifts the content around on the screen and makes it difficult to know where I am.
- [ ] What do you think about putting a "use submenu" control directly in the toolbar instead of the sidebar? I had my sidebar closed and was missing how to do that.
- [ ] Starting from the header template part in the admin isn't a great experience yet. With both sidebars open or 1 open it appears in a broken mobile view. The hamburger menu also stays for desktop mode in the admin. It would be nice to 

### Questions
- [ ] Is the masthead CTA being sticky on mobile the desired behavior? Feels a little odd to me

### General Thoughts
- [ ] It is difficult to add new menu items to the bottom menu in the editor without using the sidebar. For example, in the main UI I couldn't add a new utility nav link to Utility Item 3.

GeoffDusome (commented):
There's a whole lot of JS here from multiple different blocks that control other blocks, which is super scary to me as any change to the masthead could break any of the JS instantly.

Additionally, There's a few spots where blocks are referenced by their normal block name (`wp-block-site-title`, `wp-block-group`). Maybe we should give these custom classes so we aren't assigning styles to blocks that we may not want to? 

From the demo video: 
- Can we maybe rename the blocks within the header template part so it's obvious from a quick view what's what?
- Do we have an estimate how long it would take to edit any of this functionality to include hover instead of click. I understand this is the direction we are going, but I just KNOW we'll get a client asking if they can be hover instead and I'd like to be able to give a copy/paste answer.
- Mobile menu looks like it works great. I'm so glad we don't have to do this type of thing in every project anymore. 

I'll need to spend some time with this on a project to really understand the ins and outs and pain points or anything. 

Thanks for working on this Vinny, I realize just how much of an effort this was, even if there are time savings by using core blocks for most of it.

LayaTaal:
As we are starting to use this on 3Play one thing I am noticing is the admin experience can be pretty brutal once styles are changed to match the site. To make it easier for users we will need to adjust admin specific styles which can take quite a bit of time.

I wonder if we can setup the admin experience in such a way that it is either more agnostic to the current site design, or overall adheres more automatically to the current site design.

@GeoffDusome have you started using this yet? What has your experience been?

GeoffDusome:
@LayaTaal We haven't started using it on USC yet. We're not in a hurry for the masthead, as design hasn't been finalized, so we'll likely wait until some of the kinks have been worked out here and then implement.

Do you have any specific examples of the admin experience after styling being "brutal"? Just curious. The initial designs for the USC masthead were created before this work existed, so I'm worried that trying to throw styling onto something that wasn't even intended to be built this way in the first place will cause a bunch of issues. 

I do have concerns over the complexity for clients to update the masthead with this direction, but definitely can't think of a better way to handle this.

dpellenwood:
@Vinsanity quick reminder to add a changelog entry. ;)

wp-content/themes/core/blocks/tribe/mega-menu-item/view.js
```diff
@@ -0,0 +1,116 @@
...
+
+const el = {
+	megaMenuItems: document.querySelectorAll(
+		'.wp-block-tribe-mega-menu-item'
+	),
+};
+
+const cacheElements = () => {
+	el.masthead = document.querySelector( '.site-header' );
+};
```
GeoffDusome, line 17: 🟡 Function comment?

GeoffDusome (approved):
🟢 I'm good with this as-is, but I'm interested to know what/how we would change any of this based on usage on current projects (I don't have an example of this from my end).

PR #195, opened 2025-05-04
Update & simplify WP GA deploy workflow to WPE

lindseyolson:
## What does this do/fix?
- Updates the GA deploy workflows to WPE to use the WPE action to simplify our process

### DIFF
--- a/.github/workflows/dokku-dev.yml
+++ b/.github/workflows/dokku-dev.yml
@@ -1,5 +1,5 @@
 
-name: Dokku Deploy Dev
+name: 'Dokku Deploy Main'
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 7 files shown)


dpellenwood (commented):
@lindseyolson , Thank you for updating these!

Since we don't use WPE for every project, it probably makes more sense to move the actions so we don't have them carried around on every project.Thoughts on removing these from Moose completely and instead adding the udpated versions to Moose Resources? Maybe naming the actions `wpe-deploy-dev` etc? 

We actually even have a ticket for this in the backlog: [MOOSE-188](https://moderntribe.atlassian.net/browse/MOOSE-188) you can use that for clocking any time related to this task.

lindseyolson:
@dpellenwood Yes, I completely agree - moving them to Moose Resources is a great idea. I will add that PR on moose resources (& remove this one) later this week. Thanks for the ticket number!

dpellenwood:
@lindseyolson following up here to see if/when you might have a chance to move these?

dpellenwood (approved):
🍏

PR #198, opened 2025-05-21
Replace deprecated `::set-output`

crstauf:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


dpellenwood (approved):
🟢 Nice find, I thought I had updated all those referenced earlier. Thank you!

PR #199, opened 2025-06-03
Duplicate footer tag

LayaTaal:
## What does this do/fix?

We are already creating a footer tag in the page template. Registering another in the footer pattern creates a second, nested footer tag.


GeoffDusome (approved):

Vinsanity (approved):
🟢  TY Jason 🌮

PR #204, opened 2025-07-07
Revert post title width

LayaTaal:
## What does this do/fix?

Removes width styling on post title which is causing the title wrapper to overflow the editor screen.

## QA

Screenshots/video:

**Before**
Screenshot 2025-07-07 at 4 24 31 PM

**After**
Screenshot 2025-07-07 at 4 24 44 PM

### DIFF
--- a/wp-content/themes/core/assets/pcss/global/editor.pcss
+++ b/wp-content/themes/core/assets/pcss/global/editor.pcss
@@ -6,7 +6,6 @@
 
 /* handle styling the editor "title bar" */
 .edit-post-visual-editor__post-title-wrapper {
-	width: 100% !important;
 	margin-top: 0 !important;
 	padding-top: var(--spacer-40);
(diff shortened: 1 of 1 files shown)


dpellenwood:
> @LayaTaal I have a fix for this in #202 but it's a bit different. In that fix, I'm importing our `reset.pcss` file into the editor as elements within the editor currently aren't getting border-box box-sizing. Since the editor is now fully an `iframe` and contained, I don't see this as an issue and it fixes a whole host of additional issues that I've been seeing in editor elements that I've had to account for on USC. I'm ok with either fix, but I think I'm leaning more towards fixing the root cause. Thoughts?

Agree with Geoff here, that we're probably safe to go with the more comprehensive fix in his PR, #202 .

LayaTaal:
Nice, thank you both! Closing this now.

PR #206, opened 2025-07-07
Pattern selection modal documentation

LayaTaal:
## What does this do/fix?

Adds new documentation to help explain how to include the pattern selection modal with a new custom post type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Specific features and functionality may have additional documentation in the [./
 * [Create Block Script Templates](./docs/block-templates.md)
 * [Create WP Controls Script](./docs/wp-controls-templates.md)
 * [Supported Block Features](./docs/block-features.md)
+* [Pattern Selection Modal](./docs/pattern-picker.md)
 
 ## Modern Tribe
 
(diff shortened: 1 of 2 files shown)


dpellenwood (approved):
🍏

LayaTaal:
> 🟢 I'm fine with this, but the title of this PR is slightly misleading as the description and code in the documentation is for defining the pattern that the post should default to - not the steps to take to display the pattern selection modal. That is done via the `Block Types:` and `Post Types:` definition while defining your patterns:
> 
> ```
> Block Types: core/post-content
> Post Types: project, wp_template
> ```

I'd like to update this documentation to be a bit more clear, but honestly, it seems like

LayaTaal:
@dpellenwood @GeoffDusome I am going to close this PR.

We ran into issues on 3Play where we couldn't get the pattern picker to display even though it had the lines Geoff mentioned. Once we included the `register_block_template` code mentioned in this PR the problem was resolved. However, after testing more locally this was due to pattern caching rather than any of the code mentioned here.

I was able to fix this all finally by adding `define( 'WP_DEVELOPMENT_MODE', 'all' );` to my local-config.php. [This is described more here.](https://make.wordpress.org/core/2023/07/14/configuring-development-mode-in-6-3/).

I think we should add the development mode line to our local-config-sample.php unless there are any concerns you've already discussed. The pattern caching is a huge headache that left me and others wasting a lot of time when generating patterns and updating them.

LayaTaal:
> @LayaTaal fine with me, didn't even know that existed I always just cleared the cache.

Yet another hidden WordPress update 😄 . Flushing the cache and clearing transients doesn't work for me for patterns, so I think there is some other caching mechanism there. This seems to fully work though which is nice.

PR #210, opened 2025-07-16
Define WP_DEVELOPMENT_MODE to disable pattern caching

LayaTaal:
## What does this do/fix?

- Defines development mode constant to aide with pattern caching in local development. [See this comment for more explanation.](https://github.com/moderntribe/moose/pull/206#issuecomment-3075719887)

### DIFF
--- a/local-config-sample.php
+++ b/local-config-sample.php
@@ -48,6 +48,16 @@ if ( ! defined( 'WP_ENVIRONMENT_TYPE' ) ) {
 	define( 'WP_ENVIRONMENT_TYPE', 'local' );
 }
 
+/**
+ * Set the development mode for WordPress.
+ * Options are 'all', 'core', 'plugin', 'theme', or ''.
+ * 'all' allows us to bypass pattern caching for local development.
+ *
+ * @link https://make.wordpress.org/core/2023/10/17/improvements-to-template-loading-in-wordpress-6-4/
+ * @link https://make.wordpress.org/core/2023/07/14/configuring-development-mode-in-6-3/
+ */
+define( 'WP_DEVELOPMENT_MODE', 'all' );
+
 /*
  * Glomar
  *


GeoffDusome (approved):

local-config-sample.php
```diff
@@ -48,6 +48,16 @@
...
 
+/**
+ * Set the development mode for WordPress.
+ * Options are 'all', 'core', 'plugin', 'theme', or ''.
+ * 'all' allows us to bypass pattern caching for local development.
+ *
+ * @link https://make.wordpress.org/core/2023/10/17/improvements-to-template-loading-in-wordpress-6-4/
+ * @link https://make.wordpress.org/core/2023/07/14/configuring-development-mode-in-6-3/
+ */
+define( 'WP_DEVELOPMENT_MODE', 'all' );
```
dpellenwood, line 59: I wonder if it's sufficient for us to set the mode to `theme` or if `all` is necessary?

dpellenwood (approved):
Just a question about which value is most appropriate for our work.

local-config-sample.php
LayaTaal: @dpellenwood I did some more digging on this and I think we could probably use either, though using `all` may help avoid some edge cases where we need to do unique things in plugins. Maybe not thought.

`all` is the same as `[ 'theme', 'core', 'plugin']`, so it bypasses caching no matter if you are working on a theme, plugin, or core. While we don't work on core, there may be cases where we want this for `core` plugin work.

Here are a few references to how this work:

- [Block Pattern cache busting](https://github.com/felixarntz/wordpress-develop/blob/06f7c825842f0e55fea3b82f694b43e45d1c66df/src/wp-includes/class-wp-theme.php#L1838)
- [wp_get_development_mode](https://github.com/felixarntz/wordpress-develop/blob/06f7c825842f0e55fea3b82f694b43e45d1c66df/src/wp-includes/load.php#L333) definition

PR #211, opened 2025-07-17
[MOOSE-204] Port SVG Icon Block back to Moose Resources

spadilha:
## What does this do/fix?

Adds three versions of the Icon Picker block to the Moose Resources block library:

- **IconPicker**: Uses a custom SVG icon library.
- **IconPicker (FontAwesome)**: Uses FontAwesome icons.
- **IconPicker (MS Fabric)**: Uses Microsoft UI Fabric icons rendered via Unicode values from a custom font (`fabric-icons.woff`).

> [!IMPORTANT]
> This PR was updated to include the custom SVG version, as requested by DE.

All versions support:
- Theme or custom color palettes (for background and icon colors)
- Adjustable container size and padding
- Optional rounded shape
- Accessible labels via custom input
- Live preview of selected icon and styles

Developers can choose the appropriate version based on project needs. The MS Fabric version is ideal for Microsoft-branded environments.

> [!WARNING]
_Once approved, I’ll open a separate PR to the Moose Resources repo to add documentation._

#### General setup

1. Copy the block directories you want to use to:

```
themes/core/blocks/tribe/
```

Includes:

- `icon-picker`
- `icon-picker-font-awesome`
- `icon-picker-ms-fabric`

2. Register the blocks in `plugins/core/src/Blocks/Blocks_Definer.php`:

```php
self::TYPES => DI\add( [
  'tribe/icon-picker',
  'tribe/icon-picker-font-awesome',
  'tribe/icon-picker-ms-fabric',
] )
```

3. (Optional) In `edit.js`, switch between:
- Option 1: Use `theme.json` color palette (default)
- Option 2: Use a custom color array for icons/background

---

#### FontAwesome version (`icon-picker-font-awesome`)

If using **FontAwesome**, install the required packages:

```bash
npm install @fortawesome/react-fontawesome @fortawesome/free-solid-svg-icons
```

---

#### Custom SVG version (`icon-picker`)

1. Install SVGR CLI:

```bash
npm install --save-dev @svgr/cli
```

2. Add the following scripts to your root `package.json`:

```json
"scripts": {
  "icon-picker": "npm run icon-picker:generate-components && rm -f $npm_package_config_coreThemeBlocksDir/tribe/icon-picker/icons/components/index.js && npm run icon-picker:generate-icons-list && npm run icon-picker:strip-react-import && npm run icon-picker:lint",
  "icon-picker:generate-components": "svgr --icon --out-dir=$npm_package_config_coreThemeBlocksDir/tribe/icon-picker/icons/components $npm_package_config_coreThemeBlocksDir/tribe/icon-picker/icons/svg",
  "icon-picker:generate-icons-list": "node $npm_package_config_coreThemeBlocksDir/tribe/icon-picker/scripts/generate-icons-list.js",
  "icon-picker:lint": "wp-scripts lint-js \"$npm_package_config_coreThemeBlocksDir/tribe/icon-picker/**/*.js\" --fix",
  "icon-picker:strip-react-import": "node $npm_package_config_coreThemeBlocksDir/tribe/icon-picker/scripts/strip-react-import.js"
}
```

3. Run the following command to generate the icon components:

```bash
npm run icon-picker
```

This will:
- Convert all SVGs from `icons/svg/` into React components in `icons/components`
- Generate the `icons-list.js` file
- Lint and auto-fix the icons folder
- Strip the `import * as React` line from each generated file (SVGR adds it by default, but it's unnecessary for these simple components and triggers a linting error unless `react` is listed as a project dependency)

---

## QA

Jira ticket
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-204)

Screenshots/video:
- [Link to Video](https://www.loom.com/share/97c0c1f028084e01941dd618fe2cbb25?sid=3a202d0e-a024-4d79-8fbc-5d0c25f72829)
- [Custom SVG icons demo](https://www.loom.com/share/3d381d3b75fe44e794c7ae61bf6f5acd?sid=e79bf3b9-7442-4207-892a-c727fd254b74)


wp-content/themes/core/blocks/tribe/icon-picker/edit.js
```diff
@@ -0,0 +1,413 @@
...
+	const themeColors = useSetting( 'color.palette' );
+	const COLORS = Array.isArray( themeColors )
+		? [
+				...themeColors.map( ( { name, color } ) => ( {
+					name,
+					value: color,
+				} ) ),
+				{ name: __( 'Transparent', 'tribe' ), value: 'transparent' },
+		  ]
+		: [];
```
GeoffDusome, line 48: 🌮

GeoffDusome (approved):
🟢 I don't see anything I would change here TBH. It does everything I would want out of an icon picker block.

The one thing I can see happening here that I think would end up being a bit tough to keep up with is if design requires a custom set of icons. I completely understand how I would edit these blocks to accomplish this, but I think the part that would be hard would be managing the icon library via font family (think like icomoon). I think we would just need come up with a way with the designer on the project to update that icon set as little as possible (because who wants to update a file 20 times before launch). 

@spadilha can you think of any other way we would handle this? I would say a custom set of SVG icons but the icons would have to be built in a very specific way to be able to update the colors and such (which I think is true of services like icomoon too). 

I dunno, maybe this isn't as big of a problem as I think it is. Maybe if we can hold design to fontawesome free or the MS fabric icons in these blocks it makes this much easier, otherwise, I'd like us to round out a solution we can work together with DE on for something like this. Thoughts @dpellenwood ?

spadilha:
Thanks @GeoffDusome! I think we could easily extend this to support IcoMoon using the MS Fabric variation as a base, since it already uses a font-family to render the icons. IcoMoon also generates a `selection.json` file that we could use to list the available icons, so whenever an update is needed, we'd just need to update the font files and the JSON.

As for a set of SVG icons, we'd need a convention for how those SVGs are authored. For example, ensuring they don't have hardcoded fills or strokes and instead rely on currentColor, so we can control styling via CSS. Aside from that, I don't see any major blockers. We'd also need a consistent way to manage a list of icon names and references, which could be a little tricky to maintain, but not a big deal overall.

I believe we can stick with the FontAwesome version for now, since it already offers a good range of icons for clients. That said, we can definitely update it down the line depending on the design team's needs.

Also, upgrading to **FontAwesome Pro** would be straightforward. We'd just need to provide an access token.
Docs for reference:
https://docs.fontawesome.com/v5/web/setup/use-package-managers
https://docs.fontawesome.com/v5/web/use-with/react

spadilha:
@GeoffDusome I removed th save.js file. I also added more step-by-step instructions to the PR description, including script setup and how to generate the icons from the SVG folder. And yes, I’ll include more detailed usage instructions once we move it to the Moose Resources repo.

PR #215, opened 2025-07-21
[MOOSE-213] Core Gallery Block on Mobile

spadilha:
## What does this do/fix?

Prevent core/gallery nested images to flex-grow.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-213)


GeoffDusome (approved):
🟢 I feel like this block is rarely used in Moose, but fine with adding this in as default support regardless. Thanks!

PR #222, opened 2025-07-24
[MOOSE-245] Animation Library

GeoffDusome:
## What does this do/fix?

> [!WARNING]
> The `block-animations/index.js` file is likely collapsed below. This is the main driver file for the editing experience for the block animation library, please be sure to review that file.

> [!IMPORTANT]
> Any dynamic blocks will need to specifically add support for block animations in their render template files. While the `wp_loaded` action _does_ handle adding the attributes, the attributes are unused on dynamic blocks without specifically adding that support. 

> [!IMPORTANT]
> There are currently two places where the animation attributes, their types, and their defaults are set. Once in the `Block_Animation_Attributes` class (PHP) and once in the `attributes.js` file. If anyone has any ideas of how we can possibly avoid this that'd be great. I'd really like to avoid defining those twice. I was thinking maybe a JSON file that I could load in PHP (like we do sometimes for `theme.json`) but I have no idea where to put it.

- Adds block animation support to static & dynamic blocks included in Moose by default. 
- Enables animations on scroll via `IntersectionObserver`
- Support for adding new types of animations (ships with Fade In)
- Adds support for the dynamic PHP blocks that ship with Moose (Terms, Post Card, Search Card). 
    - Note here that the Copyright block is unaffected because it is a static block and the animation attributes get added through the normal JS script.
- Moves the Terms block helper class to the new `Helpers` directory located in the core plugin under `Blocks` 

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-245)

Screenshots/video:
- [Link to Video](https://www.loom.com/share/233c1673d09f4c29b0e190db5b9b2170)

### DIFF
--- a/wp-content/themes/core/assets/js/editor/block-animations/settings/easing.js
+++ b/wp-content/themes/core/assets/js/editor/block-animations/settings/easing.js
@@ -1,0 +1,44 @@
+/**
+ * @module easing
+ *
+ * @description pulls animation easing settings from theme.json or sets default settings
+ *
+ * theme.json settings:
+ *
+ * "animationEasing": [
+ * 		{ "label": "Ease In", "value": "ease-in" },
+ * 		{ "label": "Ease Out", "value": "ease-out" }
+ * ],
+ */
+
+import { __ } from '@wordpress/i18n';
+import themeJson from '../../../../../theme.json';
+
+const easing = themeJson?.settings?.animationEasing ?? [
+	{
+		label: __( 'Ease Out Sine', 'tribe' ),
+		value: 'cubic-bezier(0.390, 0.575, 0.565, 1.000)',
+	},
+	{
+		label: __( 'Ease In Sine', 'tribe' ),
+		value: 'cubic-bezier(0.470, 0.000, 0.745, 0.715)',
+	},
+	{
+		label: __( 'Ease In Out Sine', 'tribe' ),
+		value: 'cubic-bezier(0.445, 0.050, 0.550, 0.950)',
+	},
+	{
+		label: __( 'Ease Out Quad', 'tribe' ),
+		value: 'cubic-bezier(0.250, 0.460, 0.450, 0.940)',
+	},
+	{
+		label: __( 'Ease In Quad', 'tribe' ),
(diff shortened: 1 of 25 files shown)


wp-content/themes/core/assets/js/editor/block-animations/settings/easing.js
```diff
@@ -0,0 +1,44 @@
...
+ * ],
+ */
+
+import { __ } from '@wordpress/i18n';
+import themeJson from '../../../../../theme.json';
+
+const easing = themeJson?.settings?.animationEasing ?? [
+	{
+		label: __( 'Ease Out Sine', 'tribe' ),
+		value: 'cubic-bezier(0.390, 0.575, 0.565, 1.000)',
```
dpellenwood, line 20: Just a thought: I know we've got a PR to solve the use of this value in inline style blocks, but I wonder if it would be best to move these to global custom properties that could be used anywhere. I.e. - replace the values here with something like `--easing--ease-out-sine` and then populate the `cubic-bezier()` values in the global css var. Naming them as vars makes it easier to understand the values when they appear in code, too. ;)

wp-content/themes/core/assets/pcss/integrations/animation.pcss
```diff
@@ -0,0 +1,212 @@
...
+ * our FE IntersectionObserver in core/assets/js/utils/animate-on-scroll.js
+ * to create animations with settings on individual blocks
+ *
+ * ------------------------------------------------------------------------- */
+
+/* Setup default animation variables so they can be overridden per block */
+:root {
+	--tribe-animation-delay: 0s;
+	--tribe-animation-speed: 0.6s;
+	--tribe-animation-easing: cubic-bezier(0.39, 0.575, 0.565, 1);
```
dpellenwood, line 14: If there are global easing vars, you could use one here, too.

dpellenwood (approved):
🟢 Just one suggestion about creating a set of global easing css vars.

spadilha (approved):
LGTM! 🟢

PR #226, opened 2025-07-28
[MOOSE-239] Rating Stars block

spadilha:
Adds a custom **Rating Stars** block. It visually represents a 0–5 rating, including support for half values. The block also includes a width control (min: 100, max: 840) to adjust the size of the stars container. Developers can change these limits in `edit.js`.

- Customizable icons: replace SVG files in /icons/
- Icons are used as background-image via CSS — no need for JSX components
- Icons should be static-colored (no currentColor)
- Visual stars are decorative only and marked with `aria-hidden="true"`
- Accessible label is handled via aria-label on the container
- Icon styles are defined in `style.pcss`

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-239)

Screenshots/video:
- [Link to Video](https://www.loom.com/share/a9c2d57250e644d1b8d3826ba253819d?sid=ce3d5150-1f18-4c55-851e-36201570cf31)


wp-content/themes/core/blocks/tribe/rating-stars/render.php
```diff
@@ -0,0 +1,34 @@
...
+			if ( $remaining >= 1 ) {
+				readfile( $icons_path . 'icon-star-full.svg' );
+				$remaining -= 1;
+			} elseif ( $remaining >= 0.5 ) {
+				readfile( $icons_path . 'icon-star-half.svg' );
+				$remaining -= 0.5;
+			} else {
+				readfile( $icons_path . 'icon-star-empty.svg' );
+			}
+		}
```
GeoffDusome, line 31: ❓ I'm curious to know your thoughts on this implementation because I would have gone a different direction using like `span` tags and classes based on the rating amount (which I think is how I did it on QCHI, not that you were able to see that one), this way I was able to just add the SVGs to the global `icons.pcss` file and would use CSS's `mask` property to change the color and such. 

I'm not saying you need to change this implementation I'm just curious what the performance difference would be between this with inline SVGs as markup vs cached SVG assets. 

Any thoughts maybe @dpellenwood?

wp-content/themes/core/blocks/tribe/rating-stars/edit.js
```diff
@@ -0,0 +1,77 @@
...
+						label={ __( 'Size', 'tribe' ) }
+						value={ containerSize }
+						onChange={ ( value ) =>
+							setAttributes( {
+								containerSize: value,
+							} )
+						}
+						min={ 100 }
+						max={ 840 }
+						step={ 5 }
```
GeoffDusome, line 59: ❓ Would it be helpful to set these as constants at the top of the file instead of having to scroll down to the controls to change these values?

GeoffDusome (approved):
🟢 I don't really have any blockers here, just some discussion points. Quite a different implementation than I would have done, always cool to see a different viewpoint. 

❓ I am wondering though - @dpellenwood do you think this should be a Resources block like the Icon Picker block? I've only used rating stars on QCHI and I don't think any of our sites have used them since so they seem like something we could pull in when needed rather than having it always available, but if design needs it for basic patterns then I suppose it's fine?

wp-content/themes/core/blocks/tribe/rating-stars/edit.js
spadilha: Good catch!🌮 It's updated. Thanks!

spadilha:
@dpellenwood @GeoffDusome thanks for the feedback. I realized I was overcomplicating it. Since we don’t need to customize the SVGs in the editor, I refactored the block to use background-image spans, like Geoff mentioned. It’s much simpler now and avoids having to keep the SVGs in two places. Let me know what you think of this new version.

wp-content/themes/core/blocks/tribe/rating-stars/block.json
```diff
@@ -0,0 +1,33 @@
...
+	"description": "Displays a rating stars.",
+	"icon": "star-filled",
+	"supports": {
+		"html": false,
+		"align": false,
+		"spacing": {
+			"margin": true,
+			"padding": false
+		},
+		"__experimentalResizable": true
```
GeoffDusome, line 17: ❓ Is this being used for anything?

GeoffDusome (approved):
🟢 This feels much better than what you had before, thanks!

dpellenwood (approved):
🟢 Really clean, Saulo. Nicely done.

PR #227, opened 2025-07-29
[MOOSE-248] Logo marquee custom block

spadilha:
## What does this do/fix?

Import the Logo Marquee block created by @tylermachado into 🫎.

This PR includes two enhancements to the original block:
- Updates the block to display all logos in the editor when selected (uses a grid layout for easier editing).
- Adds drag-and-drop functionality to reorder logos directly in the editor using [dnd-kit](https://docs.dndkit.com/).

It adds two dependencies:
```
"dependencies": {
	"@dnd-kit/core": "^6.3.1",
	"@dnd-kit/sortable": "^10.0.0",
	"@dnd-kit/utilities": "^3.2.2",
	"@wordpress/icons": "^10.23.0",
	"imagesloaded": "^5.0.0"
},
```

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-248)

Screenshots/video:
- [Link to Video](https://www.loom.com/share/c6a91b07ba00496785e6a53333d10de7?sid=bca0c803-27fc-4fb3-9d71-c027b4748b47)

### DIFF
--- a/wp-content/themes/core/blocks/tribe/logo-marquee/edit.js
+++ b/wp-content/themes/core/blocks/tribe/logo-marquee/edit.js
@@ -1,0 +1,181 @@
+/**
+ * WordPress Dependencies
+ */
+
+import {
+	useBlockProps,
+	MediaPlaceholder,
+	BlockControls,
+	MediaReplaceFlow,
+} from '@wordpress/block-editor';
+import { Button, ToolbarGroup } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { useSelect } from '@wordpress/data';
+import { store as coreStore } from '@wordpress/core-data';
+import { trash } from '@wordpress/icons';
+import './editor.pcss';
+
+export default function Edit( { attributes, setAttributes, isSelected } ) {
+	const { images } = attributes;
+	const blockProps = useBlockProps();
+
+	// Fetch media details for each image
+	// eslint-disable-next-line no-unused-vars
+	const imageData = useSelect(
+		( select ) => {
+			const { getMedia } = select( coreStore );
+			return images.map( ( img ) => {
+				if ( ! img.id ) {
+					return null;
+				}
+				const media = getMedia( img.id );
+				if ( ! media ) {
+					return null;
+				}
+				return {
+					id: img.id,
+					url: img.url,
+					alt: img.alt || media.alt_text,
+					title: media.title?.raw || '',
+				};
+			} );
+		},
+		[ images ]
+	);
+
+	// Handle selection of new media items
+	const onSelectImages = ( newImages ) => {
+		setAttributes( {
+			images: newImages.map( ( image ) => ( {
+				id: image.id,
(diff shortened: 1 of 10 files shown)


wp-content/themes/core/blocks/tribe/logo-marquee/edit.js
```diff
@@ -0,0 +1,181 @@
+/**
+ * WordPress Dependencies
+ */
```
GeoffDusome, line 3: 🟡 Can probably remove this comment

wp-content/themes/core/blocks/tribe/logo-marquee/edit.js
```diff
@@ -0,0 +1,181 @@
...
+import { trash } from '@wordpress/icons';
+import './editor.pcss';
+
+export default function Edit( { attributes, setAttributes, isSelected } ) {
+	const { images } = attributes;
+	const blockProps = useBlockProps();
+
+	// Fetch media details for each image
+	// eslint-disable-next-line no-unused-vars
+	const imageData = useSelect(
```
GeoffDusome, line 24: ❓ If we need to disable eslint here, should we fix the error instead of suppressing it for the Moose implementation? I don't see `imageData` being used anywhere, can it be removed to just leave the `useSelect` by itself?

Alternatively, we should refactor this file a bit to use `withSelect` instead. You can see an example of this in the Custom Query Loop block PR in Moose Resources: https://github.com/moderntribe/moose-resources/blob/e5a1e447d08f8b64b3cb3f2c9e03d588b6b7ca1d/block-library/custom-query-loop/block/edit.js#L1088-L1146

I believe this is the intended way to handle something like this? Feel free to leave as-is though if it breaks something lol

wp-content/themes/core/blocks/tribe/logo-marquee/editor.pcss
```diff
@@ -0,0 +1,5 @@
+/**
+ * The following styles get applied inside the editor only.
+ *
+ * Replace them with your own styles or remove the file completely.
+ */
```
GeoffDusome, line 5: ❓ I noticed the editor view hides most of the images. Do we need to disable the overflow for the editor view so that the user can see all of the images added to the gallery, or do you think having the view be correct in the editor is more important than being able to actually use the "delete" function added in `edit.js` for every image?

IMO if we're going for more of a "here's a preview of what the block would be on the FE within the editor", I would remove the "delete" function from the images. Otherwise, we should fit the images in the editor into a grid instead of allowing them to overflow.

GeoffDusome (approved):
🟢 Again, no real blockers, just discussion points.

wp-content/themes/core/blocks/tribe/logo-marquee/edit.js
spadilha: Good catch @GeoffDusome 🌮! I removed the imageData hook since it’s not used anywhere and we don’t need the hook at all.

PR #231, opened 2025-08-01
[MOOSE-257] FE Linting Pre-Commit Hook

GeoffDusome:
## What does this do/fix?

- adds FE linting to pre-commit hook
    - this won't prevent some of the issues @spadilha was seeing with indenting and code editor specific things (maybe we need to look at our editor config?)
    - I specifically _chose_ to not run `--fix` on these FE hooks. I want developers to specifically see mistakes and actively correct them before committing their changes
- moved all lefthook commands to jobs to allow us to name and prevent them from running if there are no files in the commit with the types necessary to run the command
    - `phpstan` still runs on everything in the project but only runs if there are PHP changes in the commit
    - note here that any top-level file has to be called specifically as lefthook's `glob` doesn't recognize `./*.js` (etc...) as a valid selector
- updates PHP & WP versions for PHPCS 

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-257)

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -9,7 +9,7 @@ skip_output:
   - summary
 
 # code linting
-pre-commit:
+pre-push:
   parallel: true
   jobs:
     - name: phpcs
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -16,14 +16,17 @@ pre-commit:
       run: composer run phpcs
     phpstan:
       run: composer run phpstan
-    lintcss:
-      run: npm run lint:css
     lintjs:
       run: npm run lint:js
     lintconfigs:
       run: npm run lint:configs
     lintpkgjson:
       run: npm run lint:pkg-json
+  jobs:
+    - name: lintcss
+      run: npm run lint:css:fix
+      stage_fixed: true
+
 
 # prefix with jira ticket
 prepare-commit-msg:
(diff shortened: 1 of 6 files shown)


lefthook.yml
```diff
@@ -8,14 +8,47 @@ skip_output:
...
+    - name: phpstan
+      run: ./vendor/bin/phpstan analyse --memory-limit=-1
+      glob:
+        - ".phpstorm.meta.php"
+        - "local-config-sample.php"
+        - "wp-config.php"
+        - "wp-config-environment.php"
+        - "wp-content/themes/core/*.php"
+        - "wp-content/plugins/core/*.php"
+    - name: lintcss
```
dpellenwood, line 33: Thought: I wonder if it's worth adding [the formatting scripts](https://github.com/moderntribe/moose/blob/main/package.json#L55) here, too as they are what does the prettier changes (indentations, etc).

dpellenwood (approved):
I'm okay with this. Historically avoided linting on each commit as it can become an impediment to committing early & often which I'd prefer over complete & beautiful. But, I'm fine with trying it out to see how it goes.

We could alternatively do this on [a different hook](https://git-scm.com/docs/githooks#_pre_push), maybe `pre-push`? 

Just one question inline.

PR #235, opened 2025-08-20
Fix footer template-part

spadilha:
## What does this do/fix?

Fixes an issue where random menus were being displayed in the footer navigation placeholders when no specific menu was assigned. In some cases, these menus even included search and other unexpected elements. To prevent this, the `ref` value in the Navigation blocks was updated so it doesn’t auto-attach to any existing menu. Using 0, null, or similar didn’t work. The only reliable workaround was setting a high numerical value.

Also:
- Updates the footer tag.
- Fixes a typo in the copyright site name.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-156?focusedCommentId=177972)


wp-content/themes/core/parts/footer.html
```diff
@@ -1,5 +1,5 @@
...
+<!-- wp:group {"tagName":"footer","metadata":{"name":"Site Footer Pattern"},"align":"full","className":"site-footer__container alignfull is-style-dark","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"},"blockGap":"0","margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained","justifyContent":"center"}} -->
+<footer class="wp-block-group site-footer__container alignfull is-style-dark" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)"><!-- wp:group {"metadata":{"categories":["navigation","menu"],"patternName":"patterns/enhanced-nav-menu","name":"Enhanced Navigation Menu"},"align":"wide","className":"nav-menu__enhanced alignwide","style":{"spacing":{"blockGap":"var:preset|spacing|40","padding":{"right":"0","left":"0"}}},"layout":{"type":"default"}} -->
```
GeoffDusome, line 2: ❓ Did we mean to switch this back to a `footer` element? I thought this element already gets wrapped in a `footer` tag and that's why this one wasn't a footer tag. Could you double check this?

wp-content/themes/core/parts/footer.html
```diff
@@ -22,15 +22,15 @@
 <p class="is-style-default" style="margin-top:0;margin-bottom:0"><strong>Menu Title</strong></p>
 <!-- /wp:paragraph -->
 
-<!-- wp:navigation {"ref":12,"overlayMenu":"never","style":{"spacing":{"blockGap":"var:preset|spacing|20"}},"fontSize":"10","layout":{"type":"flex","flexWrap":"wrap","orientation":"vertical"}} /--></div>
+<!-- wp:navigation {"ref": 999999,"overlayMenu":"never","style":{"spacing":{"blockGap":"var:preset|spacing|20"}},"fontSize":"10","layout":{"type":"flex","flexWrap":"wrap","orientation":"vertical"}} /--></div>
```
GeoffDusome, line 25: 🟠 Alternatively to doing this, you can just remove the `ref` param all together here. It has a similar effect.

GeoffDusome (dismissed):
🔴 Can you just double check that the footer template part doesn't already get wrapped in a `footer` element?

wp-content/themes/core/parts/footer.html
spadilha: Kind of. It shows the message “Navigation Menu has been deleted or is unavailable. Create a new Menu?”
I think that’s actually the only way to get the “Create a new Menu” option to appear.

Screenshot 2025-08-21 at 17 05 35

GeoffDusome: To me, that feels like an error, rather than helpful to the editor (I guess specifically the "or is unavailable" part). Despite having an action in it, I feel like defaulting to what is essentially an error state _feels_ wrong.

On USC we ran into a lot of training issues with the Navigation block because the block itself is so complicated and convoluted that it's almost useless even though it's actually quite powerful once you get to using it. 

I'd like to wait for some input from @dpellenwood here.

dpellenwood: Ooof. Neither of these are ideal to me. The linked "Create a new menu" is so leading. I'm guessing there are already menus created at this point. So the best option would be to have it ask you to pick an existing _or_ create a new menu.

In any case, I do think having it show the error state by default is the better (of two bad) options. I've also run into the case where it randomly picks a menu and blows up the design by default and that is very jolting.

dpellenwood (approved):
🥗  Stick with the high menu ID reference value for now. It's the better of the bad options.

PR #236, opened 2025-08-21
[MOOSE-201]: Fix <details> animation

spadilha:
## What does this do/fix?

Fixes an issue where transitions on `<details>` content would not consistently trigger when toggled open and closed multiple times. This behavior was intermittent and varied between browsers and environments. On my end (macOS), the issue only happened on Brave browser. Check the demo video linked below showing the issue happening on Chrome (macOS).

### Why is this needed?

When repeatedly toggling `<details>`, browsers sometimes skip transitions due to internal optimizations (especially when no meaningful DOM/layout change is detected). This led to an inconsistent user experience, where the animation would fire the first time, then silently fail on subsequent toggles.

### What was tried (and didn’t work)?

Several CSS-only strategies were attempted, including:
- will-change: opacity, transform
- translateZ(0) to force GPU layer promotion
- Micro transition-delay variations
- pointer-events: none/auto toggling
- Using only transition instead of animation

### How does this fix it?

A small JavaScript function was added that:
- Listens for the toggle event on each `<details>` element.
- On open, it calls getComputedStyle() on each non-`<summary>` child element to force a layout reflow.
- This effectively breaks the browser’s paint-skipping optimization and ensures transitions apply every time.

This mirrors the behavior seen when inspecting elements in DevTools (where animations never fail due to forced layout).

### Notes
- The function is safe, non-invasive, and doesn’t modify styles or markup.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-201?focusedCommentId=177782)

Screenshots/video:
- [Link to Video](https://www.loom.com/share/3cb4d63024ca4a898ab3781271e76315?sid=afea18d8-a58a-4863-a003-118943b2bd4f)


dpellenwood:
> @GeoffDusome @dpellenwood I’m fine removing the animation if that’s the call, but I think we should align with the DE team first. I still need to confirm with them whether this “hack” actually solves the issue, since I couldn’t replicate it on Chrome like they mentioned. On my end, the glitch only happens on Brave, which is weird. But again, I’m totally fine pulling it.

I'm leaning towards killing the animations, too if this is what is required to make it work. It's a lot of hacking to get a relatively small UI improvement. 

I added the `Launch Environment` tag to the PR here in GitHub. If you push any change (even an empty commit) to the PR now, it should trigger an new environment to test this in using the PR as the base branch. That allows it to be tested separately, before merging the PR.

Feel free to test it out using the test env, but I'd also suggest we push back on this as a `won't fix` for the time being.

GeoffDusome (approved):

spadilha:
@GeoffDusome @dpellenwood Sarah tested it and is happy with the animation. I’m just not sure who makes the final call on whether we keep it or remove it from the project.

dpellenwood (approved):
🟢 Feel free to merge this, @spadilha . I'm fine with it. Thank you!

PR #241, opened 2025-08-22
[MOOSE-266] Color Theming

GeoffDusome:
## What does this do/fix?

> [!NOTE]
> Demo Environment: https://moose-dev-pr241.d1.moderntribe.qa/color-themes/

I'm putting together this quick draft PR to show off an alternate method to color theming that supports nested theming. Currently, we have no ability to support nested theming as the "dark" theme will always override any additional themes we add. We've seen this in every Moose site since we started doing them. 

I'm currently only supporting global stuff as well as core blocks but this would require more work as stated in my [Dev Notes comment in the ticket](https://moderntribe.atlassian.net/browse/MOOSE-266?focusedCommentId=178023).

Fixing nested theming means supporting all CSS custom properties in _all_ themes, not just overriding _some_ for themes that need it. Properties need to be able to override each other when you nest themes. 

### Why this is nice

1. Color theming is handled all in one spot. No need to go find the specific block to make the change there. Which then means...
2. No extra code within blocks for handling color theming. Theming considerations are made on individual elements as CSS custom properties and then added to the theming on the Group block level.
3. Need to add a new color theme? Copy + paste the list of properties and change values. 

### Why this kinda sucks

1. Obviously we could clean up the theme files with comments but it's a lot to take in and kinda overwhelming. 
2. Some people would consider handling theming outside of the block / element you're working on as a negative (I personally don't want to search for variables in a bunch of different places but I see your view)
3. This is a completely different way to think than we're used to in Moose currently and would require devs switch their thinking overnight. Change is hard 😞 

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-266)

Screenshots/video:
- [The current problem with nested theming](https://www.loom.com/share/e2ea247bcaf94ef2b743451a0abec508?sid=b29bf91b-d2e3-4640-b69a-e033e02b87e2)
- [What it looks when updated with the code in this PR](http://p.tri.be/i/NTew9T)


GeoffDusome:
@dpellenwood @Vinsanity @LayaTaal Could I get your eyes on this, please? I've been talking to Sarah about color theming and in order to get an editor experience that doesn't suck for specifically NESTED color theming, this is (IMO) the only way to do it. 

I'm looking to get thoughts on direction and whether or not you completely hate this before I continue the implementation. 

Thank you!

dpellenwood (commented):
Thanks for proposing this, Geoff. I'm in favor of this general approach. It makes a ton of sense to handle theming this way.

A couple things I'd be curious to explore alternatives for:
1. Would it be at all possible too not need the "default" theme at all, but instead use a fallback value within each theme's css property definition? Would that be better or worse o not possible?
2. Would it be at all possible to keep css properties associated with a specific context inside their file/partial rather than in a global themes directory? (Or maybe some combination of both?) It would be nice to not have to reference properties for a component in a totally separate location. 

Just thoughts/questions for discussion.

GeoffDusome:
Closing due to issues discussed in [this Slack thread](https://tribe.slack.com/archives/C01CY7418GY/p1758221028153569). Essentially, this method of color theming only works if we attached it specifically to a custom block that was only used in places that required color theming. It wouldn't be able to be attached to the Group block specifically, as then themes would need to be applied to each Group block down the block tree, which would get frustrating for users.

PR #256, opened 2025-09-26
[MOOSE-201] Update summary's clickable area

spadilha:
## What does this do/fix?

This pull request updates the `<details>` block by moving padding from the wrapper to the `<summary>` element, making the entire line clickable. It also shifts padding management from `theme.json` into the block’s stylesheet.

### Theme configuration updates

* Disabled `padding` support for the `core/details` block in `theme.json`, so padding is now controlled only in CSS.
* Set the default `top` and `bottom` padding for `core/details` to `0` in `theme.json` to avoid double spacing and defer spacing to CSS.

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-201?focusedCommentId=179453)

### DIFF
--- a/wp-content/themes/core/blocks/core/column/editor.js
+++ b/wp-content/themes/core/blocks/core/column/editor.js
@@ -3,12 +3,27 @@ import { __ } from '@wordpress/i18n';
 
 const settings = {
 	attributes: {
+		isSticky: {
+			type: 'boolean',
+			default: false,
+		},
 		stackingOrder: {
 			type: 'number',
 		},
 	},
 	blocks: [ 'core/column' ],
 	controls: [
+		{
+			applyClass: 'tribe-is-sticky',
+			attribute: 'isSticky',
+			defaultValue: false,
+			helpText: __(
+				'Makes column stick to top when scrolling. Only one column should be set to sticky per row and the "sticky" effect only applies on screens larger than tablet size (782px).',
+				'tribe'
+			),
+			label: __( 'Stick Column on Scroll', 'tribe' ),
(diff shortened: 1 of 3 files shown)


wp-content/themes/core/blocks/core/details/style.pcss
```diff
@@ -7,7 +7,9 @@
 		cursor: pointer;
 		list-style: none;
 		position: relative;
-		padding-right: 20px;
+		padding-right: var(--wp--preset--spacing--60);
+		padding-top: var(--wp--preset--spacing--40);
+		padding-bottom: var(--wp--preset--spacing--40);
```
GeoffDusome, line 12: 🟡 I think I'd prefer to use the theme spacing variables here:

```suggestion
		padding-right: var(--spacer-60);
		padding-top: var(--spacer-40);
		padding-bottom: var(--spacer-40);
```

wp-content/themes/core/blocks/core/details/style.pcss
```diff
@@ -26,6 +28,20 @@
...
+			text-decoration: underline;
+		}
+	}
+
+	summary + * {
+		margin-top: 0;
+	}
+
+	& > :last-child {
+		padding-bottom: var(--wp--preset--spacing--40);
```
GeoffDusome, line 44: 🟡 Same here RE: spacer variables

```suggestion
		padding-bottom: var(--spacer-40);
```

GeoffDusome (approved):
🟢 Few suggestions but should be fine as-is if you disagree :)

PR #343, opened 2026-04-08
[MOOSE-380] FE: Core Button Block "Width" Setting Support

GeoffDusome:
## What does this do/fix?

This pull request updates the Button block to ensure that when set to the "Default" style, it now properly respects the "Width" setting. The main changes involve adjusting the CSS so that button width is only set to `auto` when a custom width is not applied, improving compatibility with user-defined widths.

**Button Block Width Handling:**

* The Button block set to "Default" style now respects the "Width" setting, allowing custom widths to be applied as expected.
* In `_mixins.pcss`, the `width: auto;` property is now only applied when the button block does not have a custom width, preventing it from overriding user-defined widths.
* The direct `width: auto;` declaration was removed from the mixin to avoid unintended overrides.
* Added a comment in `style.pcss` to clarify selector specificity for overriding core styles.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-380)

Testing Environment:
- http://moose-dev-pr343.d1.moderntribe.qa/

Screenshots/video:
- Default button style respects width setting: https://p.tri.be/i/Bju45y

## Pull request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ item (Added, Changed, Depreciated, Removed, Fixed, Security).
 
 - Updated: Refactor Horizontal Tabs block into a dynamic block. Add reordering functionality. 
 - Updated: Refactor Vertical Tabs block to dynamic block; allow sorting.
+- Updated: Button block set to "Default" style now respects the "Width" setting.
 
 ## [2026.02]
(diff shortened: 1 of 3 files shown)


dpellenwood (approved):
📗

PR #323, opened 2026-02-25
fix: resolve npm peer dependency conflicts

LayaTaal:
## Summary
- Adds `overrides` block to `package.json` to pin `react` and `react-dom` to `^18.0.0`, preventing npm from resolving React 19 (via `@dnd-kit`'s loose `>=16.8.0` peer dep) which conflicts with `@wordpress/icons` and `@wordpress/scripts`
- Ported from [nexcess-2026#8](https://github.com/moderntribe/nexcess-2026/pull/8)

## Test plan
- [ ] Delete `node_modules` and `package-lock.json`, then run `npm install` — should complete without errors
- [ ] Run `npm run build` to verify blocks compile successfully

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #348, opened 2026-04-09
Chore/moose 382/package updates

dpellenwood:
## What does this do/fix?

- Update all plugins, composer & npm packages to latest
- Pin React to v18 as a peer dependency to resolve conflicts between ` @dnd-kit/core` and `@wordpress/icons`

@LayaTaal this is an alternative approach to #323 . I'm not sure which one is better, tbh. I was really hoping that WP Core would support React 19 by now, but it looks like that won't happen [until v7.1 at the earliest](https://github.com/WordPress/gutenberg/issues/71336#issuecomment-3848070272) [sigh].

## QA

Links to relevant issues
- [MOOSE-382](https://moderntribe.atlassian.net/browse/MOOSE-382)

## Pull request checklist

[MOOSE-382]: https://moderntribe.atlassian.net/browse/MOOSE-382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -54,6 +54,10 @@
 		"prettier": "npm:wp-prettier@^3.0.3",
 		"webpack-remove-empty-scripts": "^1.0.4"
 	},
+	"peerDependencies": {
+		"react": "^18.3",
+		"react-dom": "^18.3"
+	},
```
dpellenwood: Pinning React to v18 to resolve a conflict between `@dnd-kit/core` wanting v19 and and `@wordpress/icons` wanting v18.

MlKilderkin (approved):

GeoffDusome (approved):

LayaTaal (commented):
Hey @dpellenwood I was looking into this some and it looks like [@dnd-kit has a new major release](https://dndkit.com/overview) that I think will solve our problems altogether with the React version mismatch. The new version supports React 18 and 19. We will need to update our imports some to use it, though.

dpellenwood:
> Hey @dpellenwood I was looking into this some and it looks like [@dnd-kit has a new major release](https://dndkit.com/overview) that I think will solve our problems altogether with the React version mismatch. The new version supports React 18 and 19. We will need to update our imports some to use it, though.

@LayaTaal  Hmm... As I understand it, dbd-kit already supported both versions of React, but since it installed first and required v19 or v18, v19 got installed since we didn't specify, which then caused conflicts with the icons package.

Ans actually, after playing with this some more, I reverted to using overrides and added a couple more packages to resole some security warnings on install. This gives us a 100% clean `npm install`. Happy to chat about it more though.

LayaTaal:
> > Hey @dpellenwood I was looking into this some and it looks like [@dnd-kit has a new major release](https://dndkit.com/overview) that I think will solve our problems altogether with the React version mismatch. The new version supports React 18 and 19. We will need to update our imports some to use it, though.
> 
> @LayaTaal Hmm... As I understand it, dbd-kit already supported both versions of React, but since it installed first and required v19 or v18, v19 got installed since we didn't specify, which then caused conflicts with the icons package.
> 
> Ans actually, after playing with this some more, I reverted to using overrides and added a couple more packages to resole some security warnings on install. This gives us a 100% clean `npm install`. Happy to chat about it more though.

That works for me and I do think overrides makes more sense here than peer dependencies. For dnd-kit, that issue I recall is that the version we are installing had a loose React requirement of the latest version above 16, which would default to 19. But the newest version of dnd-kit specifically requires 18 or 19. The new version of dnd-kit is namespaced different than what we are importing, for example `@dnd-kit/core` has been replaced completely. My only question is, do you want to upgrade those as part of this PR or treat that as something separate?

dpellenwood:
> The new version of dnd-kit is namespaced different than what we are importing, for example `@dnd-kit/core` has been replaced completely. My only question is, do you want to upgrade those as part of this PR or treat that as something separate?

Gotcha! It's weird b/c the new version has complete changed the version numbering, too. It doesn't show as an available update when running `npm outdated`. I'll ticket that separately from this and note to come back and check the override versions as part of that ticket. Thank you!

package.json
jamiepaul: I've seen this pattern used more for React/React Native repos in open source libs than the overrides key in #323 @LayaTaal @dpellenwood but I don't know which would technically be better either.

jamiepaul (approved):
🟢 👍🏻

PR #349, opened 2026-04-09
[CONFIG] Add PostCSS support to prettier css overrides

spadilha:
## What does this do/fix?

This pull request updates the project’s Prettier configuration to properly support PostCSS files. It extends the WordPress Prettier config and adds `.pcss` to the files that should use CSS formatting rules, ensuring consistent styling across all CSS-related files.

This builds on the Prettier setup introduced in #276. The default `@wordpress/prettier-config` does not include `.pcss` in its CSS overrides, so PostCSS files were being formatted with JavaScript rules. This change addresses that gap at the project level.

**Prettier configuration improvements:**

* [`.prettierrc.js`](diffhunk://#diff-7312296367dd7c725b9beb500d7dab7bbeeb2aa03a1010a3fb2935f826d03fa6L1-R19): Now extends the WordPress Prettier config and adds `.pcss` to the list of files receiving CSS formatting rules, so PostCSS files are formatted consistently with other CSS files.

## Pull request checklist


GeoffDusome:
I'm going to let @dpellenwood have the final say here. Personally, I haven't had this issue, but it's possible I've configured my iDE to do this and haven't needed specific project configuration to do so. If this is indeed happening, I agree with what Copilot is saying here in terms of just making sure we're overriding the correct thing by checking against stuff. Entirely possible `@wordpress/prettier-config` changes and this whole thing breaks (in it's current state).

dpellenwood:
@spadilha , I've not experienced this issue.

By default prettier (via `npm run format` in ModernPress) uses WP's config, which only applies prettier to [JS & Config based files](https://github.com/WordPress/gutenberg/blob/1957df1ec31aedfe858c09478057f6a1445dd45e/packages/scripts/scripts/format.js#L106-L108). It doesn't run at all for css/pcss/sass files. 

If you run `npm run format` locally on a ModernPress site, you'll see [the output lists which files](https://p.tri.be/i/BTHTfo) are processed and pcss files are not included. 

What scenarios are you seeing where the WP Scripts format command is processing style files?

spadilha:
@dpellenwood my local setup uses Prettier as the default formatter for style files. I like it because it removes parenSpacing, switches single quotes to double quotes, cleans up spacing, and most importantly fixes indentation. Stylelint does not handle those things. How are you usually formatting CSS files?

I did not run into this in the GUQ or Mediavine projects because they do not have a `.prettierrc.js`, so I assume they were not using @wordpress/prettier-config.

In ModernPress, when I format a .pcss file locally with Prettier, it gets JS style formatting, which is the opposite of what I need. `.scss` and `.css` format correctly, so it looks like the WordPress config applies CSS rules to css, sass, and scss, but not pcss.

@GeoffDusome I changed the approach and now I am just adding a dedicated override for `.pcss`.

This got me thinking, has no one else run into this with the WordPress Prettier config and `.pcss` files? Are people generally not using Prettier for CSS formatting? If not, what are they using instead to handle indentation and other formatting that Stylelint does not cover?

GeoffDusome (approved):

dpellenwood (approved):
🍏 I'm not gonna hold this up any longer, Saulo. I still can't replicate the issues you've run into, but that's okay. Merge when ready.

PR #351, opened 2026-04-21
[MOOSE-190]: Add shared debug settings

MlKilderkin:
## What does this do/fix?

- Add `composer run sync-vscode-launch` → copies `dev/configs/.vscode/launch.json` to `.vscode/launch.json` and creates `.vscode` if missing (**overwrites** an existing `launch.json`).
- Update `docs/development-tools.md` with command, overwrite warnings, and the **PHP Debug** (`xdebug.php-debug`) requirement for VS Code

Note: There is no direct way to export PHPStorm settings for debug only. All PHP Debug servers info are stored in .idea/workspace.xml alongside other user settings
This [information](https://stackoverflow.com/questions/30349234/how-to-export-and-import-run-debug-configurations-in-phpstorm/30349440#30349440) won't work. It refers to specific debug configuration templates like running PHPUnit tests or similar

## QA

Links to relevant issues
- https://moderntribe.atlassian.net/browse/MOOSE-190


dev/scripts/inject-phpstorm-php-debug-workspace.php
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Inject or replace the PhpStorm PhpServers block in .idea/workspace.xml.
+ * Intended for Docker/Xdebug path mappings (e.g. $PROJECT_DIR$ -> /app).
+ *
+ * Run manually: composer run phpstorm-workspace-php-debug
+ *
+ * If workspace.xml already contains a different PhpServers block, asks for
+ * confirmation before overwriting. Non-interactive runs must set:
+ *   PHPSTORM_PHP_DEBUG_OVERWRITE=1
+ */
```
dpellenwood: It's a shame that the debug mappings can't be added via a shared config. I appreciate you going deep to try and automate this. But, I think this is more effort than it's worth. We can leave the documentation and a Loom demo link for how to configure PHPStorm for xDebug rather than a complex script of this nature. Imho, this is too brittle/complex to be worth keeping.

dpellenwood (commented):
@MlKilderkin , thanks for doing this. I think we should keep the simple VS Code config, but leave instructions for PHPStorm instead. Thoughts?

dev/scripts/inject-phpstorm-php-debug-workspace.php
MlKilderkin: @dpellenwood I removed script and added a link to the demo in docs. Thank you

dpellenwood (approved):
🍏 Thank you Myk!

PR #365, opened 2026-06-18
Install WP-CLI before composer install

dpellenwood:
## What does this do/fix?

We currently install WP-CLI via composer so that it's available for the WordPress installer which runs as a post-install-cmd. However, the composer-installed WP-CLI bundle has ourdated dependencies; and, it overrides any host-installed versions of WP-CLI. 

On our Dokku server, there's no way to inject the WP-CLI install _after_ PHP is installed, but _before_ composer is run as the Heroku PHP buildpack does both actions together.

Thus, I've moved the Wp-CLI install to the top of the wordpress-install script. It is conditionally run only in Dokku envs and only if WP-CLI isn't installed already.

Additionally, I've cleaned up the action versions a bit to remove any node 20 deprecation notices during deployments.

Finally, I've locked the required version of PHP to stay with 8.4. 

## Pull request checklist


igorlosouza (approved):
🟢

jamiepaul (approved):
🟢

PR #376, opened 2026-08-07
[FEATURE] Custom Faceted Directory

GeoffDusome:
## What does this do/fix?

This pull request introduces several new block controllers for faceted directory functionality, updates the block registration, and improves context handling for block rendering. The most important changes are outlined below.

---

**Faceted Directory Feature Implementation:**

* Added three new block controllers: `Directory_Grid_Controller`, `Faceted_Directory_Controller`, and `Filter_Bar_Controller` in the `Components/Blocks` namespace, providing the core logic for faceted directories, directory grid display, and filter bar controls. These controllers handle pagination, post type filtering, facet rendering, and block context management. [[1]](diffhunk://#diff-1a8e363afef83367ff699937740437709600810d4772a8f7307709c9fe0187b5R1-R152) [[2]](diffhunk://#diff-aa392083125dac230138ee7e2d4a1557b5e51bff576ca6d4ff5b6d17bfa77a50R1-R41) [[3]](diffhunk://#diff-ecba36958fed721e4bf8795e3879aab0f138225b92a632a9a20bca35ea97faf2R1-R240)

**Block Registration and Core Integration:**

* Registered new blocks (`tribe/directory-grid`, `tribe/faceted-directory`, `tribe/filter-bar`) in the `Blocks_Definer` so they are available in the editor and frontend.
* Registered new definers and subscribers for facets in the main `Core` class, ensuring the new faceted directory features are loaded and hooked into the plugin lifecycle. [[1]](diffhunk://#diff-e2f968a920a081ab1e5c68c2a8ad717f9c6fcb116cffd395d38fc8492d8e7b2eR18) [[2]](diffhunk://#diff-e2f968a920a081ab1e5c68c2a8ad717f9c6fcb116cffd395d38fc8492d8e7b2eR29)

**Block Controller Improvements:**

* Enhanced the `Abstract_Block_Controller` to support richer block context handling, including a fallback mechanism for editor preview context and a flag to control this behavior. This ensures consistent context data for both frontend and editor previews. [[1]](diffhunk://#diff-aa079454d2d7c88d95425783ec47c6bfd4088bc9f4af77bfcdaa5bd623f01440R9-R40) [[2]](diffhunk://#diff-aa079454d2d7c88d95425783ec47c6bfd4088bc9f4af77bfcdaa5bd623f01440R66-R104)

**Media Handling Enhancement:**

* Updated the `get_media` method in the `Post_Data` trait to return a placeholder image when no media is available, improving the user experience for posts without featured images.

**Changelog Update:**

* Documented the WordPress Core update, plugin, Composer, and NPM dependency updates, and ESLint config changes in the changelog.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/MOOSE-350)

Screenshots/video:
- [Link to Video](https://www.loom.com/share/269f75300cd44bf5a455ebbdfda08fa9)

Demo Environment:
- http://moose-dev-pr376.d1.moderntribe.qa/

## Pull request checklist


wp-content/plugins/core/src/Components/Blocks/Directory_Grid_Controller.php
```diff
@@ -0,0 +1,152 @@
...
+
+	/**
+	 * Total matches across every page. Prefer this over the WP_Query property:
+	 * on the indexed path WP_Query only ever sees one page.
+	 */
+	public function get_found_posts(): int {
+		return $this->found_posts;
+	}
+
+	public function get_max_num_pages(): int {
```
MlKilderkin: Do we need a getter that we never use?

wp-content/plugins/core/src/Components/Blocks/Directory_Grid_Controller.php
```diff
@@ -0,0 +1,152 @@
...
+	/**
+	 * @param array<string, mixed>|null $request
+	 */
+	private function set_query( ?array $request ): void {
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$request     ??= wp_unslash( $_GET );
+		$this->request = is_array( $request ) ? $request : [];
+		$this->paged   = max( 1, absint( $this->request[ Facet_Registry::PAGE_PARAM ] ?? get_query_var( 'paged' ) ?: 1 ) );
+
+		$built = tribe_project()->container()->get( Directory_Query::class )->build(
```
MlKilderkin, line 130: I am curios why we do in such way and don't actually do DI in constructor? `public function __construct( Directory_Query $dir_query, array $args = [] )`

wp-content/plugins/core/src/Components/Blocks/Directory_Grid_Controller.php
```diff
@@ -0,0 +1,152 @@
...
+			$this->posts_per_page,
+			$this->request,
+			$this->paged
+		);
+
+		$this->query = new \WP_Query( $built['args'] );
+
+		// The index paginates before WP_Query sees the results, so its totals
+		// are the only ones that describe the whole match set.
+		$this->found_posts = $built['total'] ?? (int) $this->query->found_posts;
```
MlKilderkin, line 141: Do we need so much extra params? We set `$this->query`. If we want the found_posts later we can use: `$this->get_query()->found_posts` or 
```
public function get_found_posts(): int {
		return (int) $this->get_query()->found_posts;
	}
```

wp-content/plugins/core/src/Facets/Directory_Query.php
```diff
@@ -0,0 +1,277 @@
+<?php declare(strict_types=1);
+
+namespace Tribe\Plugin\Facets;
+
+class Directory_Query {
```
MlKilderkin, line 5: Search and facet filtering do not compose correctly on the indexed path. The index applies facet filtering and pagination first, and only afterward does WP_Query apply the s search constraint to that already-paginated subset. total also comes from the index and therefore ignores the search query.

This can produce empty/partial pages, miss valid search results that exist on another indexed page, and show an incorrect total/page count.

I think we should either bypass the index whenever a search term is present, or make search part of the indexed page/count query so it is applied before pagination.


Example: facet has 100 matches, 10/page. Search only matches something currently in indexed page 2. Page 1 gets its 10 IDs first, then search runs against those IDs → 0 results, while total may still report 100.

wp-content/plugins/core/src/Components/Blocks/Faceted_Directory_Controller.php
```diff
@@ -0,0 +1,41 @@
...
+		parent::__construct( $args );
+
+		// block.json declares an enum, but WordPress does not enforce it at
+		// render time and this value becomes a class name.
+		$this->filter_bar_position = 'sidebar' === ( $this->attributes['filterBarPosition'] ?? '' )
+			? 'sidebar'
+			: 'top';
+
+		$post_types       = $this->attributes['postTypes'] ?? [ 'post' ];
+		$this->post_types = Facet_Registry::filter_public_post_types(
```
MlKilderkin, line 27: Could we use DI in controller instead of static method call? Same applies to other places

MlKilderkin: Or make this as trait

wp-content/plugins/core/src/Components/Blocks/Filter_Bar_Controller.php
```diff
@@ -0,0 +1,240 @@
...
+	 */
+	protected string $id_prefix;
+
+	public function __construct( array $args = [] ) {
+		parent::__construct( $args );
+
+		$position                  = (string) $this->get_context_value( self::FILTER_BAR_POSITION_CONTEXT, 'top' );
+		$this->filter_bar_position = 'sidebar' === $position ? 'sidebar' : 'top';
+		$post_types                = $this->get_context_value( self::POST_TYPES_CONTEXT, [ 'post' ] );
+		$this->post_types          = Facet_Registry::filter_public_post_types(
```
MlKilderkin, line 47: Same from above

wp-content/plugins/core/src/Facets/Directory_Query.php
```diff
@@ -0,0 +1,277 @@
...
+			'paged'          => max( 1, $paged ),
+		];
+
+		$search = isset( $request[ Facet_Registry::SEARCH_PARAM ] )
+			? sanitize_text_field( (string) $request[ Facet_Registry::SEARCH_PARAM ] )
+			: '';
+
+		if ( '' !== $search ) {
+			$args['s'] = $search;
+		}
```
MlKilderkin, line 52: By default `s` is empty. So checking its value against empty string is not necessary

`$args['s'] = isset( $request[ Facet_Registry::SEARCH_PARAM ] )
			? sanitize_text_field( (string) $request[ Facet_Registry::SEARCH_PARAM ] )
			: ''; ` is enough to cover this

MlKilderkin: Also why we don't use `$this->get_search_query()` instead?

wp-content/plugins/core/src/Facets/Directory_Query_Check.php
```diff
@@ -0,0 +1,139 @@
+<?php declare(strict_types=1);
+
```
MlKilderkin: What is the intended purpose of this file?

This looks like an ad-hoc test harness rather than production code: it lives under src/, directly requires production classes, and uses standalone assert() calls instead of the project's test infrastructure

If these cases are worth keeping, I think they should be proper automated tests under the test suite rather than a manually executable PHP file shipped with the plugin

wp-content/plugins/core/src/Facets/Facet_Index.php
```diff
@@ -0,0 +1,645 @@
...
+	private const int TERM_REINDEX_LIMIT = 500;
+
+	private bool $cache_bumped = false;
+
+	public function __construct(
+		private Facet_Registry $registry,
+	) {
+	}
+
+	public function table_name(): string {
```
MlKilderkin, line 37: Maybe it will be good to split the class into smaller once. For example move table creation to separate class

wp-content/plugins/core/src/Facets/Facet_Index.php
```diff
@@ -0,0 +1,645 @@
+<?php declare(strict_types=1);
+
+namespace Tribe\Plugin\Facets;
```
MlKilderkin, line 3: We truncate the current index while BUILT_OPTION can still be true, and it is only explicitly set to true again after the entire rebuild completes.

This may create two failure cases:

- requests during the rebuild can query a partially populated index;
- if the request times out/fatals after TRUNCATE but before the end, the partially rebuilt index can remain marked as built indefinitely.

I think the index should be marked unavailable before the destructive rebuild starts, and only marked built after the rebuild completes successfully. On failure, directory queries should continue using the taxonomy-query fallback.

wp-content/plugins/core/src/Facets/Facet_Index.php
```diff
@@ -0,0 +1,645 @@
...
+
+/**
+ * Denormalized post/term index backing faceted directory queries.
+ *
+ * One row per post + facet + term, so a multi-facet filter resolves in a single
+ * grouped query instead of one taxonomy join per facet.
+ */
+class Facet_Index {
```
MlKilderkin, line 11: Do we have kind of benchmark for this how long does it takes to rebuild the index? For example: Kinsta hosting likely fail you if your job will be run more than 30 seconds. Pantheon as well. We might need to do that in batches

wp-content/plugins/core/src/Facets/Facet_Renderer.php
```diff
@@ -0,0 +1,488 @@
...
+	/**
+	 * Mark the ancestors of every kept term, so a container parent that holds
+	 * no posts of its own cannot orphan the children underneath it.
+	 *
+	 * @param array<string, bool> $keep
+	 * @param list<\WP_Term>      $terms
+	 *
+	 * @return array<string, bool>
+	 */
+	private static function add_ancestors( array $keep, array $terms ): array {
```
MlKilderkin, line 427: Generally curios why we(AI) keep trying to use static methods? Almost all architecture is built on usage of DI and only s1 ported legacy uses static

wp-content/plugins/core/src/Facets/Facets_Subscriber.php
```diff
@@ -0,0 +1,154 @@
...
+		add_action( 'set_object_terms', function ( $object_id ): void {
+			$this->container->get( Facet_Index::class )->index_post( (int) $object_id );
+		}, 20, 1 );
+
+		// Term slug changes invalidate stored slugs for every attached post.
+		add_action( 'edited_term', function ( $term_id, $tt_id, $taxonomy ): void {
+			$this->container->get( Facet_Index::class )->reindex_term( (int) $term_id, (string) $taxonomy );
+		}, 20, 3 );
+
+		add_action( 'pre_delete_term', function ( $term_id, $taxonomy ): void {
```
MlKilderkin: pre_delete_term fires before WordPress modifies the term or its relationships, so reindex_term() calls index_post() while the term being deleted is still attached to those posts.

That means we can simply write the soon-to-be-deleted term back into the facet index and leave stale rows behind.

Could we use the post-deletion delete_term hook instead? It also provides the affected object IDs, so those posts can be reindexed after the relationships have actually been removed.

wp-content/themes/core/blocks/tribe/filter-bar/js/results.js
```diff
@@ -0,0 +1,178 @@
...
+const pushUrl = ( form, paged ) => {
+	const params = getFacetParams( form, paged ).toString();
+	const url = params
+		? `${ window.location.pathname }?${ params }`
+		: window.location.pathname;
+
+	window.history.pushState( { tribeFacets: true }, '', url );
+};
+
+export const refreshResults = async ( form, paged = 1 ) => {
```
MlKilderkin, line 121: Does this take into account next:
Filter A → request A
Filter B → request B
B finishes → correct results
A finishes → **old results overwrite B**

MlKilderkin (commented):
Could we split this into smaller PRs next time? Currently it is very hard to review

wp-content/plugins/core/src/Components/Blocks/Directory_Grid_Controller.php
GeoffDusome: I would assume it's just a style choice. I don't think there's a correct way to do this right?

wp-content/plugins/core/src/Components/Blocks/Directory_Grid_Controller.php
GeoffDusome: Saving this as an extra param is saving the total posts found, getting `found_posts` from the query would be the posts from the current page.

wp-content/plugins/core/src/Components/Blocks/Faceted_Directory_Controller.php
GeoffDusome: I don't really understand the harm in the way this is set up, if you feel strongly please feel free to make a pass at the PR.

wp-content/plugins/core/src/Facets/Facet_Index.php
GeoffDusome: Eh, I could see this as a follow-up, but it's not necessary now.

wp-content/plugins/core/src/Facets/Facet_Index.php
GeoffDusome: I have no benchmarks currently and have no real way to run any. This is a good call. I think it makes sense to look into this in a follow-up or maybe on a project that would be on a hosting platform that would require it.

wp-content/plugins/core/src/Facets/Facet_Renderer.php
GeoffDusome: I believe the idea is that the static methods have no dependency on the class itself so there's no need to use DI. I don't think static methods need to leave our codebase entirely, but maybe we need to change the location of them to a `Helpers` class or something instead?

GeoffDusome:
@dpellenwood - I'd like to give BE a chance to actually go through this work and adjust to what they think makes the most sense. Because this was written mostly by AI, it's entirely possible things can be adjusted, reordered, split out into different files, or rewritten differently (and better). 

Myk had some good points in his review, but I don't know that all of it needs to be addressed right away. Typically if we had to do something like this on a project a BE developer would have the BE created prior to me touching the FE, so we're working a bit backwards here. 

Personally, I would feel more comfortable with this if BE has a chance to actually go hands on with it to make it our own int he way we'd do it. If we could get Myk (or anyone else) some time to work on this that would be great. Because this feature is so large and important, I think it's important that we do our due diligence and do this the right way.

dpellenwood:
> @dpellenwood - I'd like to give BE a chance to actually go through this work and adjust to what they think makes the most sense. Because this was written mostly by AI, it's entirely possible things can be adjusted, reordered, split out into different files, or rewritten differently (and better).

@GeoffDusome, I agree. It's definitely worth a BE Pass. Here's [a ticket for the BE refactor work](https://moderntribe.atlassian.net/browse/MOOSE-423) on this feature. I'll get it assigned out to a BE Dev as soon as possible.

PR #384, opened 2026-08-25
Add AGENTS.md agent docs with PHP coding standards

MlKilderkin:
## What does this do/fix?

Ships the AGENTS.md + `docs/ai/` agent-instructions system, with ModernPress PHP coding standards in `docs/ai/php.md` (no separate coding-standards skill).

## Pull request checklist

- [ ] I've added a changelog entry for these changes. *(N/A unless you want a changelog for agent docs)*
- [ ] I've linked to a relevant Jira issue.
- [ ] I've captured a screenshot or screencast of the changes and linked it above. *(N/A)*


docs/ai/php.md
```diff
@@ -1,84 +1,301 @@
...
+- Do not put block logic in `render.php`. Keep templates thin.
+- Trivial attribute reads (≈1–6 lines) can stay in the template. Anything that transforms
+  attributes, branches, loops meaningfully, or computes values belongs in a controller.
+- Dynamic block controllers extend
+  `Tribe\Plugin\Components\Abstracts\Abstract_Block_Controller`; follow existing
+  controllers in `wp-content/plugins/core/src/Components/Blocks`.
+- Theme block PHP usually extends `Tribe\Plugin\Blocks\Block_Base` (core-block extensions)
+  or uses plugin controllers for dynamic `tribe/*` render templates.
+- Use `get_block_wrapper_attributes()` for wrappers. Use PHPCS escape-output ignores only
+  for intentionally safe pre-escaped markup (wrapper attrs, rendered inner blocks).
```
LayaTaal: This feels redundant. I think just a pointer in one of the two files will be enough to keep AI on track (ha!)

docs/ai/prs.md
```diff
@@ -0,0 +1,29 @@
+# Pull Requests & Commit Messages – Agent Reference
```
LayaTaal: I'm not sure we need this here. We should be using the [git-workflow skill](https://github.com/moderntribe/tribe-skills/blob/main/skills/git-workflow/SKILL.md) in tribe-skills to handle this across projects beyond ModernPress.

I would probably prefer we make updates to that skill if we need more definition.

LayaTaal (approved):
@MlKilderkin thank you for putting this together, I think it looks good. I am torn with all of the instruction in the php.md file. On the one hand I really like the very clear directions. On the other I am worried we might get force AI into situations where it doesn't know what to do. But I'm fine testing this as is for a while to see how it goes.

docs/ai/prs.md
MlKilderkin: Removed

docs/ai/php.md
MlKilderkin: Updated

MlKilderkin:
@LayaTaal I added some small changes to make it more flexible

>On the other I am worried we might get force AI into situations where it doesn't know what to do. But I'm fine testing this as is for a while to see how it goes. 

I understand the concern, but I think these constraints are useful. With lighter guidance, I have seen AI produce code that technically works but does not follow the architecture we want to maintain. That can turn a small feature into 30–60 minutes of review spent explaining architectural concerns instead of reviewing the feature itself.

Most of the rules in php.md are not arbitrary limitations. They document project decisions we already want to enforce. Without them, AI has to infer patterns from the codebase, and legacy or incorrect code can easily become precedent.

The goal is not to remove judgment, but to narrow the solution space to approaches that fit our architecture. I would rather start with clear rules and relax specific ones if they prove too restrictive in practice.

LayaTaal:
> @LayaTaal I added some small changes to make it more flexible
> 
> > On the other I am worried we might get force AI into situations where it doesn't know what to do. But I'm fine testing this as is for a while to see how it goes.
> 
> I understand the concern, but I think these constraints are useful. With lighter guidance, I have seen AI produce code that technically works but does not follow the architecture we want to maintain. That can turn a small feature into 30–60 minutes of review spent explaining architectural concerns instead of reviewing the feature itself.
> 
> Most of the rules in php.md are not arbitrary limitations. They document project decisions we already want to enforce. Without them, AI has to infer patterns from the codebase, and legacy or incorrect code can easily become precedent.
> 
> The goal is not to remove judgment, but to narrow the solution space to approaches that fit our architecture. I would rather start with clear rules and relax specific ones if they prove too restrictive in practice.

That works for me @MlKilderkin and I agree with you - we have seen together how it can not follow our architecture easily and waste our time.

LayaTaal:
@MlKilderkin once this is merged, let me know and I'll announce the base PR so the rest of the team can review.

